PR #2, opened 2018-08-16
Fixing imports from other plugins

dmaluenda:


dmaluenda:
If there is no inconvenient, I will merge this.

PR #7, opened 2018-09-17
removing deprecated code

DStrelak:
fixing:
https://sonarcloud.io/project/issues?id=Xmipp&issues=AWXOXvU5ashydt7k-Jk4&open=AWXOXvU5ashydt7k-Jk4
https://sonarcloud.io/project/issues?id=Xmipp&issues=AWXOXvU5ashydt7k-Jk5&open=AWXOXvU5ashydt7k-Jk5
https://sonarcloud.io/project/issues?id=Xmipp&issues=AWXOXvU5ashydt7k-Jk6&open=AWXOXvU5ashydt7k-Jk6
https://sonarcloud.io/project/issues?id=Xmipp&issues=AWXOXvU5ashydt7k-Jk7&open=AWXOXvU5ashydt7k-Jk7
https://sonarcloud.io/project/issues?id=Xmipp&issues=AWXOXvU5ashydt7k-Jk8&open=AWXOXvU5ashydt7k-Jk8
https://sonarcloud.io/project/issues?id=Xmipp&issues=AWXOXvU5ashydt7k-Jk9&open=AWXOXvU5ashydt7k-Jk9

PR #8, opened 2018-09-17
Ds fixing condition

DStrelak:
https://sonarcloud.io/project/issues?id=Xmipp&issues=AWXOXsvoashydt7k-IrQ&open=AWXOXsvoashydt7k-IrQ


dmaluenda:
Great!  This PR include the previous one (#7), thus we can either close the other and see both updatings as one or wrap out the c059467f572164accc2bab98a19db9a83acc22d3 commit from this PR. Even, we can accept both as they are but in order (#7 and then #8)... I think that the easiest is the first but, before fix the import in the 3Dresolution protocol. Thanks :)

rmarabini:
After refactoring I am going to review this branch. I accept all the pull request and then work with the last version

PR #21, opened 2018-11-26
Removing import as discussed in https://github.com/I2PC/scipion-em-xm…

DStrelak:
…ipp/pull/9#issuecomment-432183830


dmaluenda:
This remove has already been included in the devel branch. Sorry for late, someone else has overtaken us for the right....

PR #34, opened 2019-01-03
Bug fixes and removing Scipion protocols from the XmippViewer

dmaluenda:


dmaluenda:
Thanks a lot @yaizar!!

PR #37, opened 2019-01-11
Removed unused XmippDataset class

delarosatrevin:
As part of the pyworkflow library cleanup, the unused class Dataset was removed...so we should remove this one as well from Xmipp plugin.


dmaluenda:
Ok! Once [the Scipion's PR](https://github.com/I2PC/scipion/pull/1896) is merge, we can also merge this.

pconesa:
I didn't know that.. @dmaluenda probably @delarosatrevin is better but you might need the manager changes from this one.

dmaluenda:
As [the related Scipion's PR](https://github.com/I2PC/scipion/pull/1896) PR is merged, I will also merge this. Ok @delarosatrevin ?

delarosatrevin:
Yes, I think so. Thanks @dmaluenda

PR #42, opened 2019-01-21
adding support of the gpu alignment program

DStrelak:
Adding support for GPU version of the movie alignment.
Needs to be merged **together** with https://github.com/I2PC/xmipp/pull/27 and https://github.com/I2PC/xmippCore/pull/6 !


xmipp3/protocols/protocol_movie_correlation.py
```diff
@@ -89,22 +100,55 @@ def _defineAlignmentParams(self, form):
...
+                    expertLevel=cons.LEVEL_ADVANCED,
+                    help='Number of control points use for BSpline.')
+        line.addParam('controlPointX', params.IntParam, default=6, label='X')
+        line.addParam('controlPointY', params.IntParam, default=6, label='Y')
+        line.addParam('controlPointT', params.IntParam, default=5, label='t')
+
+        line = group.addLine('Number of patches',
+                    expertLevel=cons.LEVEL_ADVANCED,
+                    help='Number of patches to be used for patch based '
+                        'alignment. Valid only with local alignment ON.')
```
dmaluenda: If you add `condition='doLocalAlignment'` in the arguments for all the params that are only available with local alignment ON, you will not need the warning at the help anymore and Scipion will only show them if doLocalAlignment is True.

xmipp3/protocols/protocol_movie_correlation.py
```diff
@@ -174,7 +218,19 @@ def _processMovie(self, movie):
...
+        if self.useGpu.get():
+            args += ' --device %(GPU)s'
+            if self.doLocalAlignment.get():
+                args += ' --processLocalShifts '
+            args += ' --oBSpline ' + self._getExtraPath(self._getMovieRoot(movie) + "_bsplines.txt")
+            args += ' --storage ' + "fftBenchmark.txt" # assuming the protocol runs from the project folder
+            args += ' --controlPoints %d %d %d' % (self.controlPointX, self.controlPointY, self.controlPointT)
+            args += ' --patches %d %d' % (self.patchX, self.patchY)
+            args += ' --locCorrDownscale %d %d' % (self.corrDownscale, self.corrDownscale)
+            args += ' --patchesAvg %d' % self.groupNFrames
```
dmaluenda: In the help of the 'groupNFrames' parameter you warm that it's only valid with 'doLocalAlignment', but here you added anyway, it's that correct?? The same for some others above.

DStrelak:
> It would be great to still testing the non GPU version. Please, add a new test to [the movie alignment tests](https://github.com/I2PC/scipion-em-xmipp/blob/e4f16e06ba10b7d5199b3b0a48510616599a01dc/xmipp3/tests/test_protocols_xmipp_movies.py#L284). You can repeat the 'test_qbeta' but setting the `doLocalAlignment=False`.

The CPU version should behave exactly as before. New functionality is only supported in the GPU version.

xmipp3/protocols/protocol_movie_correlation.py
DStrelak: well, these are arguments for the gpu version only. The 'patches' values are then applied only when local alignment is on, otherwise they are ignored. 
But I like the conditioning you mentioned. I will add it!

xmipp3/protocols/protocol_movie_correlation.py
```diff
@@ -89,22 +100,55 @@ def _defineAlignmentParams(self, form):
...
+
         form.addParam('outsideMode', params.EnumParam,
                       choices=['Wrapping','Average','Value'],
                       default=self.OUTSIDE_WRAP,
                       expertLevel=cons.LEVEL_ADVANCED,
                       label="How to fill borders",
                       help='How to fill the borders when shifting the frames')
 
+        #GPU params
+        group = form.addGroup('GPU parameters')
```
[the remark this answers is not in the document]
DStrelak: Sounds good!

dmaluenda:
Great!

Regarding the test. Now, the default version is what is under test. In fact, I realized a contradiction on the default values for the protocol. The `useGpu=False` whereas `doLocalAlignment=True` by default and they are incompatible. I'm not sure which one should be changed... maybe set as default the new feature is more strategic (more nowadays when most of machines has GPU)...

In any case, in that moment, only the default case is under testing. Therefor, I would add a new one with the non default values. In the case that we choose `useGpu=True` by default, it's only copying the 'test_qbeta' function and
```
prot = self.newProtocol(XmippProtMovieCorr, doPSD=True, useGpu=False, doLocalAlignment=False)
```

xmipp3/protocols/protocol_movie_correlation.py
DStrelak: That is correct. 
The CPU version will not accept these new arguments (and will end up with an error if you try). 
The GPU version will accept them in any case, but as they affect the local alignment part of the computation only, they will be ignored if the 'processLocalShifts' flag is not set.

dmaluenda: :+1:

DStrelak:
>     * [ ]  The validation for the qbeta_patches test is failing due to 'You need 2 measurements per control point, i.e. use more patches in Y/X dimension or decrease number of control points.'
Yeah... the movie is little bit short, so by default there are too few control points.
I'll fix it somehow.
>     * [ ]  The CPU test is failing for me while it's not failing in the devel branch due to:
This doesn't seem to be originating in the new code. I will investigate.

DStrelak:
@dmaluenda all should be fixed now

dmaluenda:
I undersand that the close of this PR is an error from Josue. I will reopen it!

dmaluenda:
I passes the tests in carver and all is OK. If the machine don't have GPU only 1 passes and the rest fails, but I don't see a good and eassy way to avoid this.

Thus, I will say, after this scaring massive PR closings, we can merge this and [Xmipp/#43](https://github.com/I2PC/xmipp/pull/43).

DStrelak:
> Thus, I will say, after this scaring massive PR closings

:1st_place_medal:

PR #43, opened 2019-01-22
Bug xmipp ctf estimation 

JaviMota:
Bug solved when window size was bigger than the image size.


dmaluenda:
Great! It looks fine. Doesn't this imply a modification in the C++ code? If not, we can go ahead with the merge.

JaviMota:
It only affects the protocol of xmipp ctf estimation, not the C++ code

PR #48, opened 2019-02-05
Devel to master (apha->beta)

dmaluenda:
We will merge devel to master in the four Xmipp repositories (see the other pull requests), do the corresponding package (binaries bundle to Nolan) and update to pypi from the alpha version to the beta version in order to get a valid plugin to test the rest of the plugins that depens on Xmipp. Notice that this will not be the final version, but it will be useful to detect issues to solve it.


dmaluenda:
Merge this only and **only all together** with [Xmipp](https://github.com/I2PC/xmipp/pull/44), [XmippCore](https://github.com/I2PC/xmippCore/pull/11) and [XmippViz](https://github.com/I2PC/xmippViz/pull/3).

dmaluenda:
I will merge this since the plugin in pypi is updated from the alpha to the beta version.

PR #64, opened 2019-02-26
save 3D map origin 

rmarabini:
Save input volume origin in output volume. Model building programs will rejoice with change


Vilax:
I checked the changes, everything works fine. But I am not authorized to merge, if someone else can check and merge I will be gratefull

PR #73, opened 2019-03-15
Fix cudnn

rsanchezgarc:
Fix small bug installling cudnn due to new versions

### DIFF
--- a/xmipp3/__init__.py
+++ b/xmipp3/__init__.py
@@ -137,7 +137,7 @@ class Plugin(pyworkflow.em.Plugin):
 
         ## EXTRA PACKAGES ##
         # joblib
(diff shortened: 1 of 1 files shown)


dmaluenda:
I changed some things there. We should merge them to skip the conflict.

dmaluenda:
We should avoid prints and raise, because every time that `./scipion install*` is launched, the code passes to here independently if you wants to install xmipp, deepLearningToolkit or whatever

rsanchezgarc:
Sure, if there are prints, bad for me, I forgot to remove them. They are not needed at all haha.
Do I have to do anything else?

dmaluenda:
It's enough, I will test and merge it. Thx!

PR #92, opened 2019-05-06
Fi ctf fixes

fede-pe:


xmipp3/protocols/protocol_3dbionotes.py
```diff
@@ -57,25 +57,23 @@ def _insertAllSteps(self):
...
         files = {'structure_file': open(self.inputPDB.get().getFileName(), 'rb')}
 
         response = requests.post('http://3dbionotes.cnb.csic.es/programmatic/upload',data=data, files=files)
         json_data = json.loads(response.text)
-        url='http://3dbionotes.cnb.csic.es/programmatic/get/'+json_data['id']
+        url="http://3dbionotes.cnb.csic.es/programmatic/get/"+json_data['id']
         webbrowser.open_new(url)
-        fh=open(self._getExtraPath("id.txt"),"w")
-        fh.write(json_data['id'])
-        fh.close()
```
dmaluenda: This part of code don't belongs to the CTF stuff, please double check that it's correct. If don't write that file is correct, ok ;-p

dmaluenda:
How is doing this @fede-pe ? If you need some help, just tell me :)

xmipp3/protocols/protocol_3dbionotes.py
fede-pe: I am aware this has nothing to do with the CTF stuff but it is a very localized correction in a protocol (almost) nobody is using and it would be helpful if it can be included too :)

dmaluenda: Ok, I agree. Ijust concerned because it could be an accidental deleting.

dmaluenda:
We should add a help on the `refineAmplitudeContrast` parameter at https://github.com/I2PC/scipion-em-xmipp/blob/475f36a0d369094402c45e267c006fda94515018/xmipp3/protocols/protocol_ctf_micrographs.py#L115-L117 . I know this is not totally related with this PR, but we can use it to improve the readability.

Thanks!

PR #104, opened 2019-05-28
selecting gpu from form in gl2d protocols

ajimoreno:
I have included in the forms of the protocols gl2d (the three versions) the option to select the gpu but only can be used one (the program is not prepared to run in more)

### DIFF
--- a/xmipp3/protocols/protocol_classification_gpuCorr.py
+++ b/xmipp3/protocols/protocol_classification_gpuCorr.py
@@ -49,6 +49,13 @@ class XmippProtGpuCrrCL2D(ProtAlign2D):
 
     # --------------------------- DEFINE param functions -----------------------
     def _defineAlignParams(self, form):
+        form.addHidden(params.GPU_LIST, params.StringParam, default='0',
+                       expertLevel=const.LEVEL_ADVANCED,
+                       label="Choose GPU IDs",
+                       help="GPU may have several cores. Set it to zero"
+                            " if you do not know what we are talking about."
+                            " First core index is 0, second 1 and so on."
+                            " In this protocol is not possible to use several GPUs.")
         form.addParam('useReferenceImages', params.BooleanParam, default=False,
                       label='Use a Set of Reference Images ?',
                       help='If you set to *Yes*, you should provide a '
@@ -92,7 +99,7 @@ class XmippProtGpuCrrCL2D(ProtAlign2D):
                            'If *No*, all the generated classes will be '
                            'balanced',
                       expertLevel=const.LEVEL_ADVANCED,)
-        form.addParallelSection(threads=0, mpi=4)
+        form.addParallelSection(threads=0, mpi=8)
 
 
     # --------------------------- INSERT steps functions -----------------------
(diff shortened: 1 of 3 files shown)


xmipp3/protocols/protocol_classification_gpuCorr.py
```diff
@@ -49,6 +49,13 @@ class XmippProtGpuCrrCL2D(ProtAlign2D):
 
     # --------------------------- DEFINE param functions -----------------------
     def _defineAlignParams(self, form):
+        form.addHidden(params.GPU_LIST, params.StringParam, default='0',
```
DStrelak: IMHO this has nothing to do with streams. This list is a list of GPU devices to be used

xmipp3/protocols/protocol_classification_gpuCorr.py
```diff
@@ -393,13 +401,14 @@ def iterationStep (self, refSet, imgsExp, iter, useReferenceImages,
                             'keepBest': self.keepBest.get(),
                             'maxshift': self.maximumShift,
                             'outputClassesFile': filename,
+                            'device': int(self.gpuList.get()),
```
DStrelak: This is different from the line 391. Is it correct?
Also, except for the filename and outputFile, these case (flag_spit true/false) are identical. Please extract it to helper method.

xmipp3/protocols/protocol_classification_gpuCorr.py
```diff
@@ -764,6 +773,8 @@ def _validate(self):
                                   'have different sizes')
             else:
                 errors.append("Please, enter the reference images")
+        if len(self.gpuList.get())>1:
```
DStrelak: Can you use [validator](https://github.com/I2PC/scipion/blob/9baf8f71d84a62d27a701563ac09abb1e7502d9d/pyworkflow/protocol/params.py#L595) instead?

xmipp3/protocols/protocol_classification_gpuCorr.py
```diff
@@ -413,8 +422,8 @@ def iterationStep (self, refSet, imgsExp, iter, useReferenceImages,
                     self._params['Nrefs']=Nrefs
                     self._params['cl2dDir'] = self._getExtraPath(
                                               join('level%03d' % level))
-                    self.runJob("xmipp_mpi_classify_CL2D",
-                                args % self._params)
+                    self.runJob("xmipp_classify_CL2D",
+                                args % self._params, numberOfMpi=self.numberOfMpi.get())
```
DStrelak: I don't think it's necessary to pass it, but I might be mistaken

xmipp3/protocols/protocol_classification_gpuCorr.py
```diff
@@ -393,13 +401,14 @@ def iterationStep (self, refSet, imgsExp, iter, useReferenceImages,
...
                             'outputClassesFile': filename,
+                            'device': int(self.gpuList.get()),
                             }
         Nrefs = getSize(refSet)
         if Nrefs>2:
             args = '-i_ref %(imgsRef)s -i_exp %(imgsExp)s -o %(outputFile)s '\
                    '--odir %(tmpDir)s --keep_best %(keepBest)d '\
                    '--maxShift %(maxshift)d --classify %(outputClassesFile)s '\
-                   '--simplifiedMd'
+                   '--simplifiedMd --device %(device)d '
```
DStrelak: Proper format is _%(GPU)s_ (see [doc](https://scipion-em.github.io/docs/docs/developer/gpu-params-standardization.html?highlight=gpu))

DStrelak:
@ajimoreno ping

Issue #102, opened 2019-05-24
Protocols using slow version of the Fourier Reconstruction


DStrelak:
Following protocols are calling original optimized version of the Fourier Reconstruction:
- protocol_batch_cluster.py 
- protocol_projmatch\projmatch_steps.py 
- protocol_ransac.py 
- protocol_reconstruct_fourier.py 
- protocol_reconstruct_heterogeneous.py 
- protocol_reconstruct_highres.py 
- protocol_reconstruct_significant.py 
- protocol_reconstruct_swarm.py 
- protocol_split_volume_hierarchical_cluster.py 
- protocol_validate_overfitting.py 

We should somehow rework them, so that they can use the optimized CPU/GPU version, and, where applicable, the approximate (faster, but less precise) version of it.

PR #109, opened 2019-05-30
Replacing original implementation of the Fourier reconstruction by th…

DStrelak:
…e accelerated one (and GPU one)

Adressing  #102

Everywhere (except batch_cluster) has been added support of GPU and threads
Highres is using only non-approximative version.
Reconstruct_fourier, split_volume and heterogeneous have approximative version by default, but that can be changed.
Other protocols are using approximative version only.

Please **test your plugin**, as I am not sure how to do it properly.

I tested Highres, overfitting and reconstruct_fourier.

### DIFF
--- a/xmipp3/protocols/nma/protocol_batch_cluster.py
+++ b/xmipp3/protocols/nma/protocol_batch_cluster.py
@@ -32,7 +32,6 @@ from pyworkflow.em.data import SetOfParticles, Volume
 from xmipp3.convert import writeSetOfParticles
 
 
-
 class BatchProtNMACluster(BatchProtocol):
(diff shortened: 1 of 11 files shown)


DStrelak:
> Maybe we should edit the help of the GPU section in order to explain that the GPU will be used only for the Fourier transforms...

I think that this is unnecessary.
As a user, I would prefer to simply tell Scipion that this and that hardware is available, and leave it up to it to process my data. I am not really interested if particular program will run on CPU or GPU, I just want it to run fast and correctly :)

DStrelak:
Basically, to keep the protocols simple.

PR #110, opened 2019-05-31
remove line  outVol.setObjLabel(vol.getObjLabel())

rmarabini:
outVol.setObjLabel(vol.getObjLabel()) asigns to the new volume exactly the same label than the input volume has. In this way they are two different objects with the same alias and therefore, there is no way to distinguish them. I did not remove the line:

            outVol.setObjComment(vol.getObjComment())

but I also think it should be gone

### DIFF
--- a/xmipp3/protocols/protocol_align_volume.py
+++ b/xmipp3/protocols/protocol_align_volume.py
@@ -189,7 +189,6 @@ class XmippProtAlignVolume(em.ProtAlignVolume):
         for vol in self._iterInputVolumes():
             outVol = em.Volume()
             outVol.setLocation(vol.outputName)
-            outVol.setObjLabel(vol.getObjLabel())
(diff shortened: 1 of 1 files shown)


dmaluenda:
I agree. Which label will be set as default? Maybe we can do something like
```
outVol.setObjLabel("%s (aligned)" % vol.getObjLabel())
```

For the comment I would leave as it is. But I don't know, I've never used the comments....

rmarabini:
Just deleting the line is OK. The default label, that contains the protocol
name, usually is good enougth.

       Roberto

DStrelak:
@rmarabini Please rebase and merge this PR

dmaluenda:
I merged this PR. @rmarabini delete this branch if you will not use it anymore. I leave it since its name is quite generic and, perhaps, you want to keep it.

PR #115, opened 2019-06-12
devel << pypiVersion

dmaluenda:


dmaluenda:
@Vilax, have a look on this. Here there are the changes that we included to the release. Do you want this changes in devel or you are working on this separatelly?

If you doubt, the best option is to include this changes in devel. Because in this way, both branches will be totally coherent.

Thanks!!

xmipp3/protocols/protocol_resolution_monotomo.py
```diff
@@ -90,13 +91,17 @@ def _defineParams(self, form):
                       help='The mask determines which points are specimen'
                       ' and which are not')
 
+        #form.addParam('wedge', BooleanParam, default=False,
```
DStrelak: remove if no longer needed

xmipp3/protocols/protocol_resolution_monotomo.py
```diff
@@ -90,13 +91,17 @@ def _defineParams(self, form):
...
+        #              label="Weight resolution by wedge?", 
+        #              help='A second local resolution map is provided,'
+	#		'by filtering the local resolution map by the missing'
+	#		'wedge.')
+
         group = form.addGroup('Extra parameters')
 
         line = group.addLine('Resolution Range (Å)',
                             help="If the user knows the range of resolutions or"
```
DStrelak: 'user' is the one reading it, so I think it's better to address him/her directly

xmipp3/protocols/protocol_resolution_monotomo.py
```diff
@@ -117,23 +122,24 @@ def _defineParams(self, form):
...
                                   }
         self._updateFilenamesDict(myDict)
 
     def _insertAllSteps(self):
             # Convert input into xmipp Metadata format
         self._createFilenameTemplates() 
-        self._insertFunctionStep('convertInputStep', )
-        self._insertFunctionStep('resolutionMonogenicSignalStep')
+        self._insertFunctionStep('convertInputStep')
+	#if self.wedge.get():
```
DStrelak: remove if no longer needed

xmipp3/protocols/protocol_resolution_monotomo.py
```diff
@@ -200,7 +206,12 @@ def maskRadius(self):
         return xdim
     
 
-    def resolutionMonogenicSignalStep(self):
+    def detectMissingWedgeStep(self):
```
DStrelak: Wedge is removed, no? Is this method used from somewhere?

xmipp3/protocols/protocol_resolution_monotomo.py
```diff
@@ -214,38 +225,41 @@ def resolutionMonogenicSignalStep(self):
         params = ' --vol %s' % self.vol1Fn
         params += ' --vol2 %s' % self.vol2Fn
         params += ' --meanVol %s' % self._getFileName(FN_MEAN_VOL)
+
+	#if self.wedge.get():
+	#    params += ' --filteredMap %s' % self._getFileName(FN_FILTERED_MAP)
```
DStrelak: remove

xmipp3/protocols/protocol_resolution_monotomo.py
```diff
@@ -214,38 +225,41 @@ def resolutionMonogenicSignalStep(self):
...
         params += ' --significance %f' % self.significance.get()
-        params += ' --md_outputdata %s' % self._getFileName(METADATA_MASK_FILE)  
 
         self.runJob('xmipp_resolution_monotomo', params)
 
 
 
     def createHistrogram(self):
 
+        if self.stepSize.hasValue():
```
DStrelak: you can use freq_step = self.stepSize.get() if self.stepSize.hasValue() else 10

xmipp3/viewers/viewer_resolution_monogenic_signal.py
```diff
@@ -252,57 +258,98 @@ def createChimeraScript(self):
         fninput = abspath(fnbase + ext[0:4])
 
         dim = inputVolume.getDim()[0]
-        tmpFileName = os.path.abspath(self.protocol._getTmpPath("axis.bild"))
-        Chimera.createCoordinateAxisFile(dim,
+
+	imageFile = self.protocol._getFileName(OUTPUT_RESOLUTION_FILE_CHIMERA)
+
+        if not os.path.exists(imageFile):
+            if self.protocol.halfVolumes.get() is True:
```
DStrelak: 'if sth is True' is just longer way of saying 'if sth' :)

xmipp3/viewers/viewer_resolution_monogenic_signal.py
```diff
@@ -252,57 +258,98 @@ def createChimeraScript(self):
...
+                smprt = self.protocol.inputVolume.get().getSamplingRate()
+            else:
+                smprt = self.protocol.inputVolumes.get().getSamplingRate()
```
DStrelak: Consider renaming the fields.
inputVolume vs inputVolumeS is a very subtle different. It will cause hard to find bugs.

xmipp3/viewers/viewer_resolution_monogenic_signal.py
```diff
@@ -252,57 +258,98 @@ def createChimeraScript(self):
...
+
+            imageFile = replaceExt(imageFile, 'vol')
+            fhCmd.write("open %s\n" % fninput)
+	    fhCmd.write("open %s\n" %  abspath(imageFile))
```
DStrelak: Wrong indent.
Python is not complaining about this?

xmipp3/viewers/viewer_resolution_monogenic_signal.py
```diff
@@ -252,57 +258,98 @@ def createChimeraScript(self):
...
+	    fhCmd.write("volume #0 voxelSize %s\n" % (str(smprt)))
+            fhCmd.write("volume #1 voxelSize %s\n" % (str(smprt)))
+            fhCmd.write("vol #1 hide\n")
+        
+            scolorStr = '%s,%s:' * numberOfColors
+            scolorStr = scolorStr[:-1]
```
DStrelak: Why (are you removing the colon at the end)? Is it not better to remove it from the 'template' above?

xmipp3/viewers/viewer_resolution_monogenic_signal.py
```diff
@@ -252,57 +258,98 @@ def createChimeraScript(self):
...
+
+            line = ("scolor #0 volume #1 perPixel false cmap " 
+                + scolorStr + "\n") % colorList
+            fhCmd.write(line)
+
+            scolorStr = '%s %s ' * numberOfColors
+            str_colors = ()
+            for idx, elem in enumerate(colorList):
```
DStrelak: This loop looks strange. Sometimes you are using str(elem), sometimes not, sometimes you are putting elem on odd position, sometimes not ....

xmipp3/viewers/viewer_resolution_monogenic_signal.py
```diff
@@ -252,57 +258,98 @@ def createChimeraScript(self):
...
                     % (imageFileSmprt, x_output, y_output, z_output))
 
-        #### Check if the coordinate system works for a set of volumes
+            #### Check if the coordinate system works for a set of volumes
 
-        fhCmd.write("volume #2 hide\n")
-        
-        scolorStr = '%s,%s:' * numberOfColors
-        scolorStr = scolorStr[:-1]
+            fhCmd.write("volume #2 hide\n")
```
DStrelak: Same as code above. Extract this to a method and it instead.

xmipp3/viewers/viewer_resolution_monotomo.py
```diff
@@ -108,8 +108,8 @@ def _defineParams(self, form):
         group.addParam('doShowVolumeColorSlices', LabelParam,
               label="Show colored resolution slices")
         
-        group.addParam('doShowVolumeColorFiltered', LabelParam,
-              label="Show colored filtered resolution slices")
+        #group.addParam('doShowVolumeColorFiltered', LabelParam,
```
DStrelak: remove

xmipp3/viewers/viewer_resolution_monotomo.py
```diff
@@ -175,6 +175,29 @@ def _showVolumeColorSlices(self, mapFile):
         xplotter.getColorBar(plot)
 
         return [plt.show(xplotter)]
+
+    def _showVolumeColorSlicesTwoMaps(self, mapFile1, mapFile2):
+        imageFile = self.protocol._getFileName(mapFile1)
+        imgData, min_Res, max_Res = self.getImgData(imageFile)
+	imageFile = self.protocol._getFileName(mapFile2)
```
DStrelak: wrong indent

DStrelak:
@dmaluenda ping

dmaluenda:
Same that for https://github.com/I2PC/xmipp/pull/161, @Vilax is in charge.

xmipp3/viewers/viewer_resolution_monotomo.py
dmaluenda: fixed

xmipp3/protocols/protocol_resolution_monotomo.py
```diff
@@ -91,17 +86,11 @@ def _defineParams(self, form):
...
-	#		'wedge.')
-
         group = form.addGroup('Extra parameters')
-
         line = group.addLine('Resolution Range (Å)',
-                            help="If the user knows the range of resolutions or"
-                                " only a range of frequencies needs to be analysed.")
+                            help="Resolution range (and step) to evaluate "
+                                 "the local resolution. If step is missing, "
+                                 "0.25 is used.")
```
DStrelak: Don't mention default value, as should it change in the code, but not here, it would create inconsistencies.
Use see the default value in the protocol.

xmipp3/viewers/viewer_resolution_monogenic_signal.py
DStrelak: I understand, but like this it's very easy to create a new bug. 
If you use two variables - halfVolume/entireVolume, this would be less likely.
Other option would be to call it 'inputVol', and then check the type of it at runtime

xmipp3/viewers/viewer_resolution_monogenic_signal.py
DStrelak: What I meant is that the blocks are equal, i.e. they should be extracted to a single function/method

xmipp3/viewers/viewer_resolution_monogenic_signal.py
dmaluenda: Well, we usually use this syntax. Singular for a EMobject and plural for a EMset... Now, we have released this protocol using `inputVolume` and `inputVolumes` attribute names, then we should to keep this naming in order to keep the backwards compatibility when a project from the old version is opened with the new version.

xmipp3/viewers/viewer_resolution_monogenic_signal.py
dmaluenda: I don't get you. No code is repeated now. The equal blocks are merged together after the conditional.

xmipp3/protocols/protocol_resolution_monotomo.py
dmaluenda: Done

PR #119, opened 2019-06-26
use right gold volume in reconstruction tests

pconesa:


xmipp3/tests/test_convert_xmipp.py
```diff
@@ -140,7 +140,7 @@ def launchTest(self, fileKey, mList, alignType=None, **kwargs):
             goldFn = self.dataset.getFile(fileKey + '_Gold_output.mrcs')
         else:
             outputFn = self.getOutputPath(fileKey + "_output.vol")
-            goldFn = self.dataset.getFile(fileKey + '_Gold_output.vol')
+            goldFn = self.dataset.getFile(fileKey + 'Gold')
```
dmaluenda: Great. I've spent a lot of time this morning trying to find the error on this test. This is fixing it, my question is why this have changed?

pconesa: No idea why this suddenly start failing.

PR #124, opened 2019-07-10
Rb carbon cleaner

rsanchezgarc:
Bug fixed in continue protocol
Bug fixed in multigpu

### DIFF
--- a/xmipp3/protocols/protocol_deep_micrograph_screen.py
+++ b/xmipp3/protocols/protocol_deep_micrograph_screen.py
@@ -364,16 +364,6 @@ class XmippProtDeepMicrographScreen(ProtExtractParticles, XmippProtocol):
     def notOne(self, value):
         return abs(value - 1) > 0.0001
 
-    # def _getNewSampling(self):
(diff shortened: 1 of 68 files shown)


xmipp3/protocols/protocol_deep_micrograph_screen.py
```diff
@@ -217,7 +217,7 @@ def _computeMaskForMicrographList(self, micList):
             if self.useQueueForSteps() or self.useQueue():
                 args += ' -g all '
             else:
-                args += ' -g %(GPU)s'
+                args += ' -g %s'%(",".join([str(elem) for elem in self.getGpuList()]))
```
DStrelak: are you sure this is the right way?

rsanchezgarc: that is the reason I changed it. micrographCleaner is expecting as -g a list of comma separated gpu ids. E.g.

-g 1,2

And previously, it was receiving

-g 1 2

So in the end, it was just making use of one gpu.

PR #129, opened 2019-07-15
Enabling empty mics in consensus picking

dmaluenda:
When some micrograph was not picked for some of the input pickers, the consensus was entering in an infinity loop waiting for that mic.

Now, if streaming is active, only those mics ready (non empty) for all the inputs are processed and, then, when streaming finishes, the rest of the mics are processed.

### DIFF
--- a/xmipp3/protocols/protocol_particle_pick_consensus.py
+++ b/xmipp3/protocols/protocol_particle_pick_consensus.py
@@ -94,8 +94,8 @@ class XmippProtConsensusPicking(ProtParticlePicking):
 #--------------------------- INSERT steps functions ---------------------------
     def _insertAllSteps(self):
         self.inputMics = 0
-        self.check = 0
-        self.mics = []
+        self.processedMics = []
+        self.checkedMics = set()
         self.setOfCoords = []
         self.inputs = self.inputCoordinates[0].get()#min([coor.get() for coor in ])
         coorSteps = self.insertNewCoorsSteps([])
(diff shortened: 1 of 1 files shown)


delarosatrevin:
I haven't looked at the code in detail...but I'm not sure if the self.processedMics (and related lists/dicts) is populated when the protocol continues (i.e. in case it has been stopped/continued). I think it should be check when the protocol is run in continue mode, that all micrographs in the outputset are put initially in this self.processedMics.

pconesa:
@delarosatrevin, I never think about that case....good to know.

dmaluenda:
You right @delarosatrevin . I've just pushed an update taking this in account:

Now, if self.checkedMics is empty, we look for files in the extra to populate it and self.processedMics.
If you go from a clean run, extra is empty and nothing else.
If it's from a stop/continue, we assume that all processed mics has his posFile in extra.

Additionally, I updated the way to populate the outputSet. Sometimes it was not closing it. Now, I think it's clear.

xmipp3/protocols/protocol_particle_pick_consensus.py
```diff
@@ -167,141 +178,83 @@ def _checkNewInput(self):
...
+                    coords = [coords]
+                for coord in coords:
+                    newCoord = Coordinate()
+                    micrographs = self.mainInput.getMicrographs()
+                    newCoord.setMicrograph(micrographs[getMicId(fnTmp)])
+                    newCoord.setPosition(coord[0], coord[1])
+                    outSet.append(newCoord)
+
+            outputName = 'consensusCoordinates'
+            firstTime = True
```
DStrelak: firstTime = False if self.hasAttribute(outputName) else True

xmipp3/protocols/protocol_particle_pick_consensus.py
```diff
@@ -167,141 +178,83 @@ def _checkNewInput(self):
...
                     self._defineTransformRelation(inCorrds, outSet)
             outSet.close()
+
         if self.finished:  # Unlock createOutputStep if finished all jobs
             outputStep = self._getFirstJoinStep()
             if outputStep and outputStep.isWaiting():
-                outputStep.setStatus(cons.STATUS_NEW)
-        else:
-            return
+                outputStep.setStatus(STATUS_NEW)
```
DStrelak: if you finish, and you're waiting, you se status to new? Sounds strange

xmipp3/protocols/protocol_particle_pick_consensus.py
dmaluenda: Even better: `firstTime = not self.hasAttribute(outputName)`

xmipp3/protocols/protocol_particle_pick_consensus.py
dmaluenda: It's sounds strange, but this is how Scipion's streaming is designed...

First you create a `createOutputStep`, the you set dependecies for it (all new Steps) and, finally, you set the `createOuputStep` as NEW to launch it if all deps are ready...

xmipp3/protocols/protocol_particle_pick_consensus.py
```diff
@@ -314,3 +267,81 @@ def _summary(self):
...
+        consensus = consensus.get()
+    consensusCoords = allCoords[votes >= consensus, :]
+    try:
+        jaccardIdx = float(len(consensusCoords)) / (
+                float(len(allCoords)) / Ninputs)
+        # COSS: Possible problem with concurrent writes
+        with open(jaccFn, "a") as fhJaccard:
+            fhJaccard.write("%d %f\n" % (getMicId(posFn), jaccardIdx))
+    except:
+        pass
```
DStrelak: cmon, at least report an error

xmipp3/protocols/protocol_particle_pick_consensus.py
dmaluenda: Well, when `createOutputStep` is created, it's set to a WAITING status to avoid be launched when it's dependencies are finished, but more data is coming but not arrived yet.

xmipp3/protocols/protocol_particle_pick_consensus.py
dmaluenda: jajaja, this is how it was... I just copied ;-p

I've just updated

PR #134, opened 2019-07-26
Use pointers with extended in test

pconesa:
If we link a protocol directly to an output like: 

prot.inputParticles.set(ptotImport.outputParticles)

It will work but copy, export to json will lose the link. It is better to do:

prot.inputParticles.set(ptotImport)
prot.inputParticles.setExtended("outputParticles")

Or as in this PR, Create a pointer to a protocol and set the extended.

### DIFF
--- a/xmipp3/protocols/protocol_movie_average.py
+++ b/xmipp3/protocols/protocol_movie_average.py
@@ -30,6 +30,7 @@ import os
 
 import pyworkflow.protocol.params as params
 from pyworkflow import VERSION_1_1
+from pyworkflow.utils import getExt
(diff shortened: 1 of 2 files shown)


dmaluenda:
I totally agree! However, the bast majority of tests are using `prot2.parameter1.set(prot1.output)` instead of `prot2.parameter1.set(prot1)` + `prot2.parameter1.setExtended('output')`. Therefore, we should review them. If not, we are testing a different think then in real processing.

pconesa:
I think both approaches are present at the moment. We should recommend, even maybe not accept tests without the "good" approach from now on.

PR #143, opened 2019-08-30
Extracting as MRC instead of STK saves data conversions downstream, e…

cossorzano:
….g., calling CryoSparc


azazellochg:
I wish we could avoid using stk at all costs in all xmipp protocols, like mask creation, preprocess etc. But I guess I'm "anti-xmipp" fan :D

delarosatrevin:
@azazellochg I think xmipp can deal also with mrc stacks as well as spider stack, so this will not be an issue...and I agree to use mrcs by default if this will save conversions to move into other programs.

PR #144, opened 2019-09-03
Fix deep consensus dot

rsanchezgarc:
Bug fixing nasty fnames

### DIFF
--- a/xmipp3/protocols/protocol_screen_deepConsensus.py
+++ b/xmipp3/protocols/protocol_screen_deepConsensus.py
@@ -528,12 +528,14 @@ class XmippProtScreenDeepConsensus(ProtParticlePicking, XmippProtocol):
                 if baseName not in inputCoordsFnames:
                     inputCoordsFnames[baseName]=["None"]*nCoordsSets
                 inputCoordsFnames[baseName][coord_num]= os.path.join(tmpPosDir, posFname)
-      inputFileStr="#pos_i\n"
+      inputFileHeader="#pos_i\n"
+      inputFileStr=inputFileHeader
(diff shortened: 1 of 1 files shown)


xmipp3/protocols/protocol_screen_deepConsensus.py
```diff
@@ -528,12 +528,12 @@ def calculateCoorConsensusStep(self, outCoordsDataPath, mode):
...
                     inputCoordsFnames[baseName]=["None"]*nCoordsSets
                 inputCoordsFnames[baseName][coord_num]= os.path.join(tmpPosDir, posFname)
-      inputFileStr="#pos_i"
+      inputFileStr="#pos_i\n"
       for baseName in inputCoordsFnames:
          fnames= inputCoordsFnames[baseName]
          inputFileStr+=" ".join(fnames)+"\n"
              
-      assert len(inputFileStr)>0, "Error, no consensus can be computed as there are mismatch in coordinate sets filenames"
+      assert len(inputFileStr)>8, "Error, no consensus can be computed as there are mismatch in coordinate sets filenames"
```
dmaluenda: It's a little bit dirt... I've spend about 10minutes to understand why `8` here. I would use an auxiliary variable like
```
inputFileInit = "#pos_i"
inputFileStr = inputFileInit

for ...

assert len(inputFileStr)-len(inputFileInit) > 0, "Error ... "
```
or
```
assert len(inputFileStr.lstrip(inputFileInit)) > 0, " Error ... "
```

rsanchezgarc: You are right. This was a very first patch and I forgot to improve it.

PR #147, opened 2019-09-09
fix micro_cleaner continue

rsanchezgarc:
Fixed bug in micro_cleaner continue

### DIFF
--- a/xmipp3/protocols/protocol_deep_micrograph_screen.py
+++ b/xmipp3/protocols/protocol_deep_micrograph_screen.py
@@ -106,13 +106,12 @@ class XmippProtDeepMicrographScreen(ProtExtractParticles, XmippProtocol):
                            "Manual thresholding can be performed after execution through analyze results button. "+
                            "\n0.75 <= Recommended threshold <= 0.9")
 
-        form.addParam("streamingBatchSize", params.IntParam, default=36,
+        form.addParam("streamingBatchSize", params.IntParam, default=50,
                       label="Batch size", expertLevel=params.LEVEL_ADVANCED,
                       help="This value allows to group several items to be "
                            "processed inside the same protocol step. You can "
                            "use the following values: \n"
-                           "*0*    Put in the same step all the items "
-                           "available."
+                           "*0*    Put in the same step all the items  available.\n "
                            "*>1*   The number of items that will be grouped into "
                            "a step.")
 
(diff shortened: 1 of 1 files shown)


xmipp3/protocols/protocol_deep_micrograph_screen.py
```diff
@@ -106,13 +106,12 @@ def _defineParams(self, form):
                            "Manual thresholding can be performed after execution through analyze results button. "+
                            "\n0.75 <= Recommended threshold <= 0.9")
 
-        form.addParam("streamingBatchSize", params.IntParam, default=36,
+        form.addParam("streamingBatchSize", params.IntParam, default=50,
```
dmaluenda: Me parece un valor demasiado alto. Piensa que durante un procesamiento en streaming las movies llegan con un ritmo del order de 1 movie por minuto. Esto se traduciría con que se trabajaría con batches de 50minutos. Es decir que el workflow estará parado 50min para luego liberar una gran cantidad de información. Yo iría por un valor de alrededor de 10.

xmipp3/protocols/protocol_deep_micrograph_screen.py
```diff
@@ -197,31 +196,33 @@ def _computeMaskForMicrographList(self, micList):
...
-            if self.useQueueForSteps() or self.useQueue():
-                args += ' -g all '
-            else:
-                args += ' -g %s'%(",".join([str(elem) for elem in self.getGpuList()]))
-        else:
-            args += ' -g -1'
-            
-        self.runJob('xmipp_deep_micrograph_cleaner', args)
+
+        if len(micLisfFn)>0:
```
dmaluenda: ok

xmipp3/protocols/protocol_deep_micrograph_screen.py
rsanchezgarc: What I would really want to do is:

If protocol is running in streaming:
   batchSize= 8
else:
  batchSize= numberOfMicrographs

Take in mind that every time you call a deep learning program, it has to load the model to gpu and it takes a lot of time.

Is it possible to detect if a protocol is running in streaming or in static mode?

Ruben

delarosatrevin: Ruben, even if the protocol is not in streaming, I still like to set a batchSize > 1, so you still seeing results quicker than waiting fro the whole set of micrographs to be processed. This is quite convenient to detect possible parameters problems even if the protocol is not in streaming. 

Just for the record, there is also the option 0, that means to take all movies that are available. So, combined with the sleepTime, can work pretty similar to what you like.

delarosatrevin: I agree with @dmaluenda that 50 seems quite high value...but I don't know how much it does take to load the model and libraries compared to processing the data.

rsanchezgarc: Loading models is a lot. It is not worth it at all to process less than 16 micrographs. It is like 10 seconds loading and 4 seconds computing.

The key here would be to  autodetect if running in streaming  and then select an appropriate default threshold (it could be indicated with batchSize=-1)

If you always want to see some first results, but no performance degradation desired, another alternative would be to execute always a first small batch and them bigger batches

delarosatrevin: I think this is quite a bit too much work for not much outcome. I think batchSize=0 will just do that...take the total amount of available input items. Anyway, you could also have the convention for your protocol that -1 means compute the optimum value, but I guess the same does not apply to all protocols in streaming.

rsanchezgarc: Ok. Although I think that this should apply to all protocols using tensorflow.

dmaluenda: In streaming 10s + 4s/mic is not too long and a batch size of 10 would be enough.

I mean, using batches of 10mics (which means a batch computation of 50s) is enough even for a very fast acquisitions of 10s/movie (resulting in 100s to fill each batch), so it will stay a half of the time waiting for filling the batch and processing during the other half. For slower acquisitions, it stay more time waiting than processing.

Instead, using batches of 50mics, they will be processed in 3.5mins and it will wait for 5min more for fill the batch. Since in non-streaming processing the performance is highly increasing, I would vote for a dynamic behavior depending of if its in streaming or not.

You can use the [protocol.isInStreaming()](https://scipion-em.github.io/docs/modules/pyworkflow/protocol/protocol.html#Protocol.isInStreaming) method to check the mode that is running. However, it should be check after launch the protocol, if not its inputs are not defined yet...

rsanchezgarc: Ok.

I will try to use  protocol.isInStreaming() solution

PR #169, opened 2019-11-29
Dhg cv2

DaniDelHoyo:
This is not a true pull request since there are two files (__init__.py and protocol_consensus_clustering.py that slipped into the commit). David, could you have a look at the movie_gain? It is fixing two problems: 1) the summary was calculated always on the same gain image; 2) the output set did not have all the input movies but only those for which the gain was calculated


xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -46,67 +46,6 @@
...
-def array_zeros_to_median(a, thres=0.01, depth=1):
-    '''Return an array, replacing the zeros (values under a threshold) with the median of
-    its surrounding values (with a depth)'''
-    idxs = np.where(np.abs(a) < thres)[0]
-    idys = np.where(np.abs(a) < thres)[1]
-
-    for i in range(len(idxs)):
-        sur_values = surrounding_values(a, idxs[i], idys[i], depth)
-        a[idxs[i]][idys[i]] = np.median(sur_values)
-    return a
```
dmaluenda: I send all this functions to the end of the file

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -176,8 +115,7 @@ def _insertNewMoviesSteps(self, insertedDict, inputMovies):
                 deps.append(stepId)
                 insertedDict[movie.getObjId()] = stepId
         else:
-            if (len(insertedDict) == 0 and self.estimateOrientation.get()
-                and self.inputMovies.get().getGain() is not None):
+            if len(insertedDict) == 0 and self.estimateOrientation.get():
```
dmaluenda: The `self.inputMovies.get().getGain() is not None` assertion now is done in the validate() function.

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -195,72 +133,54 @@ def _insertNewMoviesSteps(self, insertedDict, inputMovies):
                 self.convertCIStep.append(normStepId)
 
             # For each movie insert the step to process it
-            for movie in inputMovies:
-                if movie.getObjId() not in insertedDict:
-                    stepId = self._insertMovieStep(movie)
-                    deps.append(stepId)
-                    insertedDict[movie.getObjId()] = stepId
+            for idx, movie in enumerate(self.inputMovies.get()):
```
dmaluenda: The enumerate is not needed anymore. We should take off in order to increase performance. I forgot to delete it.

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -195,72 +133,54 @@ def _insertNewMoviesSteps(self, insertedDict, inputMovies):
...
     def estimateOrientationStep(self, movieDict):
         movie = Movie()
-        movie.setAttributesFromDict(movieDict, setBasic=True,
-                                    ignoreMissing=True)
+        movie.setAttributesFromDict(movieDict, setBasic=True, ignoreMissing=True)
+
         movieId = movie.getObjId()
-        fnMovie = movie.getFileName()
-        gain = self.inputMovies.get().getGain()
-        args = "-i %s --oroot %s --iter 1 --singleRef --frameStep %d" % \
```
dmaluenda: this args are used more than one time, so I've made a function

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -195,72 +133,54 @@ def _insertNewMoviesSteps(self, insertedDict, inputMovies):
...
-               (fnMovie, self._getPath("movie_%06d" % movieId),
-                self.frameStep)
-        fnGain = self._getPath("movie_%06d_gain.xmp" % movieId)
+        movieFn = movie.getFileName()
+        expGainFn = self.inputMovies.get().getGain()
+        resGainFn = self.getCurrentGain(movieId)
 
         # Check which estimated gain matches with the experimental gain
-        if self.estimateOrientation.get() and gain is not None and not os.path.exists(
-                self._getPath("bestGain.xmp")):
```
dmaluenda: this assertion is made in the validation() function

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -195,72 +133,54 @@ def _insertNewMoviesSteps(self, insertedDict, inputMovies):
...
+        resGain.read(resGainFn)
+        expGain = xmippLib.Image()
+        expGain.read(expGainFn)
+        self.match_orientation(expGain, resGain)
 
-            if fnGain is not None:
-                oriGain = xmippLib.Image()
-                oriGain.read(fnGain)
-                oriArray = oriGain.getData()
+    def normalizeGainStep(self):
```
dmaluenda: Here is now the nomalizeGainStep(), is just spliting the old one.

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -195,72 +133,54 @@ def _insertNewMoviesSteps(self, insertedDict, inputMovies):
...
         args = self.getArgs(fnMovie, movieId)
-        fnGain = self.getCurrentGain(movieId)
 
-        if self.useExistingGainImage.get() and gain is not None:
-            if self.estimateOrientation.get() or self.normalizeGain.get():
-                args += " --gainImage %s" % self.getBestGain()
-            else:
-                args += " --gainImage %s" % gain
+        if self.useExistingGainImage.get() and inputGain is not None:
+            args += " --gainImage %s" % self.getFinalGain()
```
dmaluenda: getFinalGain() is taking the best or the original if best doesn't exist.

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -269,16 +189,18 @@ def _processMovie(self, movie):
...
-            #No previous gain: bestGain is the estimated
-            if gain is None:
-                os.system("cp %s %s" % (self._getPath("movie_%06d_gain.xmp" % movieId),self._getPath("bestGain.xmp")))
-            #Previous gain: bestGain is the original
+        # If the gain hasn't been oriented or normalized, we still need bestGain
+        if not os.path.exists(self.getBestGain()):
+            # No previous gain: bestGain is the estimated
+            if inputGain is None:
+                pass
+                # copyFile(self.getCurrentGain(movieId),
```
dmaluenda: I think it's not needed. However, in the case that it's needed, copyFile() function from pwutils module is more clear.

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -489,20 +412,44 @@ def getCurrentGain(self, movieId):
         return self._getExtraPath("movie_%06d_gain.xmp" % movieId)
 
     def getBestGain(self):
-        return self._getExtraPath("bestGain.mrc")
+        return self._getExtraPath("bestGain.xmp")
```
dmaluenda: my fault! .mrc is better than .xmd. If you could change it, it will be great!

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -489,20 +412,44 @@ def getCurrentGain(self, movieId):
...
         return ("-i %s --oroot %s --iter 1 --singleRef --frameStep %d %s"
                 % (movieFn, self._getExtraPath("movie_%06d" % movieId),
-                self.frameStep, extraArgs))
+                   self.frameStep, extraArgs))
 
     def doGainProcess(self, movieId):
         return movieId == 1 or movieId % self.movieStep.get() == 0
 
     # --------------------------- INFO functions -------------------------------
+    def _validate(self):
```
dmaluenda: validate() function is called when user click "Execute". Then if some incompatibility in the form parameters is found, a message is raised with the returned list items.

DaniDelHoyo:
In principle, this should be the final version

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -432,7 +430,7 @@ def getArgs(self, movieFn, movieId, extraArgs=''):
                    self.frameStep, extraArgs))
 
     def doGainProcess(self, movieId):
-        return movieId == 1 or movieId % self.movieStep.get() == 0
+        return (movieId-1) % self.movieStep.get() == 0
```
dmaluenda: good point!

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -465,30 +463,6 @@ def _summary(self):
 
 # --------------------- WORKERS --------------------------------------
 
-def cv2_applyTransform(imag, M, shape):
```
dmaluenda: Is this removing intentionally?

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -465,30 +463,6 @@ def _summary(self):
...
-    ''' Apply a transformation(M) to a np array(imag) and return it in a given shape
-    '''
-    (hdst, wdst) = shape
-    transformed = cv2.warpAffine(imag, M[:2][:], (wdst, hdst),
-                                 borderMode=cv2.BORDER_CONSTANT, borderValue=1.0)
-    return transformed
-
-
-def cv2_rotation(imag, angle, shape, P):
```
dmaluenda: Is this removing intentionally?

DaniDelHoyo:
Hi David, yes I removed them intentionally because they were the functions that used the cv2 module which we decided not to use.

DaniDelHoyo:
the alternatives that use a xmipp function for the transformation were already added

PR #170, opened 2019-12-03
Protocol_connected_components

EstrellaFG:
This protocol takes a set of coordinates and identifies connected components among the picked particles.


xmipp3/protocols/protocol_connected_components.py
```diff
@@ -0,0 +1,132 @@
...
+# *
+# *  All comments concerning this program package may be sent to the
+# *  e-mail address 'scipion@cnb.csic.es'
+# *
+# **************************************************************************
+
+import numpy as np
+from pyworkflow.em.protocol import EMProtocol
+from pyworkflow.protocol.params import PointerParam, FloatParam
+from tomo.protocols import ProtTomoBase
```
dmaluenda: In order to avoid exceptions when plugin not found, please use this to import from other plugin
```
from pyworkflow.utils import importFromPlugin
ProtTomoBase = importFromPlugin("tomo.protocols", "ProtTomoBase")

EstrellaFG: Solved here: https://github.com/I2PC/scipion-em-xmipp/pull/170/commits/dd9d6885e1b9f3f26bf8ec26d6c66254fb113848

PR #171, opened 2019-12-04
Remove duplicates created

cossorzano:


dmaluenda:
It's not able to work in streaming, right? It would be nice if so..

cossorzano:
It would certainly be. But I don't know how to make protocols in streaming

dmaluenda:
I'm working in the streaming version in this branch and I think it will be ready soon. Thus, let me finish before merging this PR.

dmaluenda:
I've refactorized the ConsensusPicking protocol in order to inherit from it and use it as base.

PR #173, opened 2019-12-05
Co consensus picking

cossorzano:


xmipp3/protocols/protocol_particle_pick_consensus.py
```diff
@@ -343,7 +345,11 @@ def consensusWorker(coords, consensus, consensusRadius, posFn, jaccFn):
         consensus = Ninputs
     else:
         consensus = consensus.get()
-    consensusCoords = allCoords[votes >= consensus, :]
+    if mode==0:
```
DStrelak: can this be done differently (enum, maybe)? This approach is error-prone

xmipp3/__init__.py
```diff
@@ -58,7 +58,7 @@ def getEnviron(cls, xmippFirst=True):
         environ.update({
             'PATH': getXmippPath('bin'),
             'LD_LIBRARY_PATH': getXmippPath('lib'),
-            'PYTHONPATH': getXmippPath('pylib')
+            'PYTHONPATH': getXmippPath('pylib')+":"+getXmippPath(os.path.join('pylib','xmippPyModules'))
```
dmaluenda: I don't agree on this. Probably the module must be load by
```
from xmippPyModules.moduleName import blablabla
```
or similar.

xmipp3/protocols/protocol_particle_pick_consensus.py
cossorzano: I agree, I have added some labels in the same way as in the rest of protocols

xmipp3/__init__.py
cossorzano: I cannot remember in which context this error appeared, but this line was the only way we found to solve it.

PR #192, opened 2020-01-20
Devel fixes

dmaluenda:
- Cleaning swarm consensus
- Bug fixing of extract-particles test: We changed the extract borders default value in the protocol. In this PR, the tests is modified according.

### DIFF
--- a/xmipp3/tests/test_protocols_xmipp_mics.py
+++ b/xmipp3/tests/test_protocols_xmipp_mics.py
@@ -506,12 +506,14 @@ class TestXmippExtractParticles(TestXmippBase):
         self.assertTrue(particle.hasAttribute('_xmipp_scoreByVariance'),
                         'Particle has not scoreByVariance attribute.')
         self.assertAlmostEqual(particle._xmipp_scoreByVariance.get(), varianceScore,
-                               3, "The was a problem with the varianceScore")
+                               delta=0.01,
+                               msg="The was a problem with the varianceScore")
 
         self.assertTrue(particle.hasAttribute('_xmipp_scoreByGiniCoeff'),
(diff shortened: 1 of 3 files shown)


xmipp3/tests/test_protocols_xmipp_mics.py
```diff
@@ -591,16 +593,16 @@ def compare(objId, delta=1.0):
         self._checkSamplingConsistency(outputParts)
         self._checkVarianceAndGiniCoeff(outputParts[170], 1.2081, 0.5754)
 
-    def testNoExtractBorders(self):
-        print "Run extract particles avoiding extract in borders"
+    def testExtractBorders(self):
+        print "Run extract particles extracting borders particles"
```
DStrelak: What is it testing?
It seems that before it was testing the particles, which were not near some border.
Now it's testing all particles? Particles near border?

dmaluenda: Before, the protocol was set by default in such a way that particles at border were extracted and the missing part is full it in some way. Then, this test was checking the non default behavior (non extracting the particles at borders).

Now, the protocol do NOT extract particles at border by default. Then, this test as it was doesn't make sense because is the same than the others (the default behavior). Thus, now is rewritten in order to check the non default behavior (extracting borders particles) by checking one border particle existence (335) and its stats.

DStrelak: So wouldn't it be better to print `Extract particles at border`?

dmaluenda: mmm, ok. Why not..

PR #193, opened 2020-01-27
devel fixes

dmaluenda:


xmipp3/protocols/protocol_movie_max_shift.py
```diff
@@ -214,10 +214,12 @@ def fillOutput(newDoneList, firstTime, AccOrDisc='Accepted'):
             micsDwSet = self._loadOutputSet(SetOfMicrographs,
                                   'micrographs_dose-weighted%s.sqlite' % suffix)
 
-            def tryToAppend(outSet, micOut, tries=1, label='movie'):
+            def tryToAppend(outSet, micOut, tries=1, labelTmp='movie (%s)'):
```
DStrelak: how about sth like:
`labelPrefix = 'movie' `
`labelStr = ' '.join(labelPrefix, micOut.getMicName())`

I don't feel comfortable to pass a string which is expected to have some specific formatiing.

dmaluenda: Agree, done!

PR #197, opened 2020-02-18
Merging devel in python3_migration

dmaluenda:
This is to take all the updated code from devel.

I've run all the tests getting more failing ones than in buildbot/devel but **less than in buildbot/sdevel** (which is testing python3_migration branch). Indeed, no new fails should appear after this PR, but at the opposite, some of them will be fixed.

Tu summarize the failing errors, they are as follow:
 - Due to OF (no openCV installed)
 - Due to DeepLearningToolkit not installed
 - Due to xmipp_cuda_reconstruct_fourier (I think that the problem is `--device %(GPU)s` is passed literally)
 - ConvertPdb and ConvertToPseudoAtoms protocols don't exist
 - and some others that I don't property understand


DStrelak:
This PR is **huge**. 
Please try to avoid this in future.
What is the difference between  `python3_migration` and `devel_python3`?

dmaluenda:
I know.. the `python3_migration` is the branch that @fonsecareyna82 did al september in order to migrate the plugin to python3. Now this is concluded but `devel` has been recolecting updates from september to now. Then, this bring this updates from `devel` (in python2) to this branch while updating them to python3.

So, `devel_python3` collect those changes marged to `devel` (migrated to python3) and after this PR, python3_migration and devel will be identical but one in python2 and the other in python3.

xmipp3/__init__.py
```diff
@@ -213,18 +213,17 @@ def tryAddPipModule(env, moduleName, *args, **kwargs):
 def installDeepLearningToolkit(plugin, env):
     deepLearningTools = []
 
+    scikit_image = tryAddPipModule(env, 'scikit-image', '0.14.2',
```
pconesa: Why not using standard reqirements.txt?

xmipp3/__init__.py
```diff
@@ -213,18 +213,17 @@ def tryAddPipModule(env, moduleName, *args, **kwargs):
...
+                                   target='scikit_image*',
+                                   default=False, deps=['scipy', 'scikit-learn'])
+    deepLearningTools.append(scikit_image)
+
     # pandas
     pandas = tryAddPipModule(env, 'pandas', '0.20.1',
```
pconesa: and in all cases replace tryAddPipModule by adding line is requirements.txt? Although we will loose the chance of being defaults =False ...unmm

xmipp3/convert/convert.py
```diff
@@ -219,13 +222,13 @@
...
     ALIGNMENT_DICT = OrderedDict([
-           ("_xmipp_shiftX", emlib.MDL_SHIFT_X),
-           ("_xmipp_shiftY", emlib.MDL_SHIFT_Y),
-           ("_xmipp_shiftZ", emlib.MDL_SHIFT_Z),
-           ("_xmipp_flip", emlib.MDL_FLIP),
-           ("_xmipp_anglePsi", emlib.MDL_ANGLE_PSI),
-           ("_xmipp_angleRot", emlib.MDL_ANGLE_ROT),
-           ("_xmipp_angleTilt", emlib.MDL_ANGLE_TILT),
+           (prefixAttribute("shiftX"), emlib.MDL_SHIFT_X),
+           (prefixAttribute("shiftY"), emlib.MDL_SHIFT_Y),
```
pconesa: I like this.

xmipp3/protocols/protocol_3dbionotes.py
```diff
@@ -56,6 +56,7 @@ def _defineParams(self, form):
     
     #--------------------------- INSERT steps functions ------------------------
     def _insertAllSteps(self):
+        self.url = String("")
```
pconesa: I think this protocol can be deleted, there a plugin for it.

xmipp3/protocols/protocol_movie_max_shift.py
```diff
@@ -215,33 +215,55 @@ def fillOutput(newDoneList, firstTime, AccOrDisc='Accepted'):
...
+                    micOut.setEnabled(enable)
+                    outSet.append(micOut)
+                except Exception as ex:
+                    if tries < 10:
+                        from time import sleep
+                        sleep(tries*3)
+                        tryToAppend(outSet, micOut, tries+1)
+                    else:
+                        labelStr = ' '.join([labelPrefix, micOut.getMicName()])
+                        self.warning("The %s seems corrupted. Skkiping it...\n "
```
pconesa: Skiping

xmipp3/utils.py
```diff
@@ -145,3 +146,51 @@ def validateDLtoolkit(errors=None, **kwargs):
 
     return errors
 
+BAD_IMPORT_TENSORFLOW_KERAS_MSG='''
+Error, tensorflow/keras is probably not installed. Install it with:\n  ./scipion installb deepLearningToolkit
+If gpu version of tensorflow desired, install cuda 8.0 or cuda 9.0
+We will try to automatically install cudnn, if unsucesfully, install cudnn and add to LD_LIBRARY_PATH
+add to SCIPION_DIR/config/scipion.conf
```
pconesa: Use here the variable for the config file: pw.Config.SCIPION_CONFIG

xmipp3/__init__.py
dmaluenda: Well, this will change soon because I'm working on the virtualenvs.

Notice that if they are included in the requirements, they will coexist will Scipion modules that can be incompatible (as we have seen before).

If your question is why this change here, the answer is because it is like in devel ;-p

xmipp3/__init__.py
dmaluenda: Same as above, it will be in a virtualenvironment

delarosatrevin: > and in all cases replace tryAddPipModule by adding lines is requirements.txt? Although we will lose the chance of being defaults =False ...unmm

I think there is a way in the setup.py to specify optional dependencies...I would suggest to try to use the standards methods of installation and specifying deps (boths in setup.py and in requirements.txt)

dmaluenda: As commented above, I'm working on launching python codes (inside the Xmipp package) under virtual environments in order to avoid collisions with Scipion modules. Thus, this change is here because it is coming from devel, but all that pip installations will disappear soon and they will go in the virtual environment definition.

Regarding the use of the standard method, I think that we should put in the requirements.tx or setup.py only that modules needed by the plugin, but not those modules that the software behind needs (let's say keras, tensorflow, scikit-learn, pandas...) due to some of them has incompatibilities with certain versions of matplotlib and other basic modules that Scipion uses.

xmipp3/protocols/protocol_movie_max_shift.py
dmaluenda: Even 'skipping' XD. Thanks!

xmipp3/protocols/protocol_3dbionotes.py
dmaluenda: Done

xmipp3/utils.py
dmaluenda: Done

xmipp3/__init__.py
pconesa: I see, then is fine. It will be much aseptic later.

xmipp3/protocols/protocol_movie_max_shift.py
pconesa: yep...

dmaluenda:
Hi there,

@DStrelak have suggested to go with this via `git rebase` in order to avoid this kind of massive-commits PR. After some reading and tries, I think it's a go idea. It's as follow:
```
Before rebase:
          A---B---C      python3_migration
         /
    D---E---...---G      devel

After rebase:
                    A'--B'--C'   python3_migration
                   /
    D---E---...---G              devel
```

I've tried and all seems pretty correct. The tests passes as in this PR (devel_python3 branches) and the merging have gone quite smooth in this way. So, I will close these PRs and I will push the rebase.

What do you think about it? @pconesa @fonsecareyna82 @delarosatrevin @cossorzano

delarosatrevin:
@dmaluenda I think the suggested rebase solution is pretty common and useful for avoiding a lot of small commits and having a cleaner history.

dmaluenda:
I've reopen this PR because I'm not comfortable with the rebase approach.

PR #198, opened 2020-03-04
Update protocol_volume_local_sharpening.py

erneyramirez:
eliminate symmetry


dmaluenda:
Ok. Please, do the same in the `python3_migration` branch.

Tip: since it is a very simple change, you can do it from the GitHub GUI.

PR #199, opened 2020-03-05
Regarding bindings and installation

dmaluenda:
In this PR has been included the following:

 - Merging duplicated code in `scipion-em-xmipp/xmipp3/base.py` and `bindings/python/xmipp-base.py` (see [#244 in xmipp repo](https://github.com/I2PC/xmipp/pull/244)).

 - Linking the bindings and the `libXmipp.so` and `libXmippCore.so` into the Scipion's directories after any Xmipp installation (xmippSrc, xmippDev, xmippBin_Centos, xmippBin_Debian).

 - Allowing to install (also re-compile) by means of `scipion3 installb xmippDev`.

 - Xmipp dependencies have been removed (commented), they will be either in requirements.txt or under a virtual environment (the same for the deepLearningToolkit)

### DIFF
--- a/xmipp3/__init__.py
+++ b/xmipp3/__init__.py
@@ -125,31 +125,32 @@ class Plugin(pwem.Plugin):
             Scipion-defined software can be used as dependencies
             by using its name as string.
         """
-        # Cancel for now all binaries installation
-        return
 
+        # scons = tryAddPipModule(env, 'scons', '3.0.4')
+        # joblib = tryAddPipModule(env, 'joblib', '0.11', target='joblib*')
+        #
+        # # scikit
+        # scipy = tryAddPipModule(env, 'scipy', '1.4.1', default=True)
+        # cython = tryAddPipModule(env, 'cython', '0.29.14', target='Cython-0.29*',
+        #                          default=True)
+        # scikit_learn = tryAddPipModule(env, 'scikit-learn', '0.22',
+        #                                target='scikit_learn*',
+        #                                default=True)
 
-
-        scons = tryAddPipModule(env, 'scons', '3.0.4')
-        joblib = tryAddPipModule(env, 'joblib', '0.11', target='joblib*')
-
-        # scikit
-        scipy = tryAddPipModule(env, 'scipy', '1.4.1', default=True)
-        cython = tryAddPipModule(env, 'cython', '0.29.14', target='Cython-0.29*',
-                                 default=True)
-        scikit_learn = tryAddPipModule(env, 'scikit-learn', '0.22',
-                                       target='scikit_learn*',
-                                       default=True)
-
-        xmippDeps = [scons, joblib, scikit_learn]
+        xmippDeps = []  # scons, joblib, scikit_learn]
         ## XMIPP SOFTWARE ##
         lastCompiled = "lib/libXmippJNI.so"
-        targets = [cls.getHome('bin', 'xmipp_reconstruct_significant'),
-                   cls.getHome(lastCompiled)]
+        installTgt = [cls.getHome('bin', 'xmipp_reconstruct_significant'),
+                      cls.getHome(lastCompiled)]
 
         compileCmd = ("src/xmipp/xmipp config && src/xmipp/xmipp check_config && "
                       "src/xmipp/xmipp compile %d && touch DONE && rm -rf %s 2>/dev/null"
                       % (env.getProcessors(), cls.getHome()))
+        compileTgt = ["src/xmippViz/"+lastCompiled, "DONE"]
+
+        installCmd = "rm DONE ; src/xmipp/xmipp install %s" % cls.getHome()
+        xmippBashrc = cls.getHome('xmipp.bashrc')
+        verToken = cls.getHome('v%s' % _currentVersion)
 
         # TODO: Check that getHome returns an absolute path
         bindingsAndLibsCmd = ("ln -s %s %s ; ln -s %s %s"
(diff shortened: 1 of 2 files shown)


xmipp3/__init__.py
```diff
@@ -125,207 +125,231 @@ def defineBinaries(cls, env):
...
+        xmippDeps = []  # scons, joblib, scikit_learn]
         ## XMIPP SOFTWARE ##
         lastCompiled = "lib/libXmippJNI.so"
-        targets = [cls.getHome('bin', 'xmipp_reconstruct_significant'),
-                   cls.getHome(lastCompiled)]
+        installTgt = [cls.getHome('bin', 'xmipp_reconstruct_significant'),
+                      cls.getHome(lastCompiled)]
 
         compileCmd = ("src/xmipp/xmipp config && src/xmipp/xmipp check_config && "
                       "src/xmipp/xmipp compile %d && touch DONE && rm -rf %s 2>/dev/null"
```
DStrelak: Last time I have seen code like this (in my previous work), our senior developer lost his entire home folder :D
Honestly, I don't understand what is happening here, but this whole method looks scary.
Why does `defineBinaries` clone and build the repo? This to me is a huge side effect.

xmipp3/__init__.py
```diff
@@ -125,207 +125,231 @@ def defineBinaries(cls, env):
...
+                                 cls.getHome('lib', 'libXmippCore.so'),
+                                 Config.SCIPION_LIBS))
+        bindingsAndLibsTgt = ([os.path.join(Config.SCIPION_BINDINGS, 'xmipp_base.py'),
+                               os.path.join(Config.SCIPION_BINDINGS, 'xmippLib.so'),
+                               os.path.join(Config.SCIPION_LIBS, 'libXmipp.so')])
+
+        pluginDir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+        if (os.path.basename(os.path.dirname(pluginDir)) == 'src'  # xmipp-bundle
+            and os.path.isfile(os.path.join(pluginDir, 'setup.py'))):  # devel mode
+            develBranch = os.environ.get('XMIPP_DEVEL_BRANCH', None)
```
pconesa: Not a default one?

pconesa: What is the scenario here...the user has cloned just xmipp plugin? therefore you clone the others.
Probably we can assume 2 scienarios here:

a: developer has only clone xmipp-plugin ---> therefore the others have to be cloned.
b: developer has clone xmipp "standard way"...so we skip the clone since all should be there

?

dmaluenda: Right, it should be NOT by default.

Indeed, this is a remanent code that I did when we consider the A scenario. However, as we have seen some days ago, it doesn't make much sense. Thus, I'm modifying this in order to avoid de cloning part but allowing to rebuild the xmipp-bundle thru the plugin.

Then, the scenario that I'm considering is:
1st step: developer installs scipion+xmipp in devel mode thru the pypi.scipion-installer
2nd step: developer codes something in any Xmipp-C++ repo
3rd step: developer compile that by using `scipion3 installb xmippDev`

If you think that this scenario (especially the 3rd step) doesn't make sense, I might agree. I'm not 100% sure of the utility of this, rather about 50%, since the 3rd step can be done just invoking `./xmipp` in a terminal by the developer...

My pros is: An external developer has xmipp-bundle there, he/she can code and compile Xmipp knowing nothing about `./xmipp` command, just dealing with `scipion3` command.

My cons is: More complexity, more maintenance.

Bets open ;-p

pconesa: I like this, not for xmipp developers, but for other em developers. This should simplify things for them.

xmipp3/__init__.py
```diff
@@ -125,207 +125,231 @@ def defineBinaries(cls, env):
...
+        compileTgt = ["src/xmippViz/"+lastCompiled, "DONE"]
+
+        installCmd = "rm DONE ; src/xmipp/xmipp install %s" % cls.getHome()
+        xmippBashrc = cls.getHome('xmipp.bashrc')
+        verToken = cls.getHome('v%s' % _currentVersion)
+
+        # TODO: Check that getHome returns an absolute path
+        bindingsAndLibsCmd = ("ln -fs %s %s ; ln -fs %s %s ; ln -fs %s %s"
+                              % (cls.getHome('bindings', 'python', '*'),
+                                 Config.SCIPION_BINDINGS,
```
pconesa: Due to recent changes in pyworkflow thes contants are no longer accesses like this. There are 2 methods:
Config.getLibFolder() and Config.getBindingsFolder()

xmipp3/__init__.py
```diff
@@ -34,7 +34,7 @@
 import pyworkflow.utils as pwutils
 
 from .base import *
-from .constants import XMIPP_HOME
+from .constants import XMIPP_HOME, XMIPP_URL
```
dmaluenda: @pconesa check this please. I think it was a bug.

I will take this to ask you for review this PR again ;-p.

Thanks!!

xmipp3/__init__.py
dmaluenda: I've re-considered the way how we can allow developers (outside xmipp team) re-compiling xmipp without passing thru xmipp bundle, just `scipion3`command or, even, the plugin manager.

Now, the `defineBinaries` just invoke `./xmipp` with the `all` flag; while, the cloning is made by the [scipion-installer](https://pypi.org/project/scipion-installer/) when `-dev` flag is passed (except if `-noXmipp` flag is passed too)

xmipp3/__init__.py
pconesa: It's not a bug, I've added a constant xmipp_url and its use at setup.py and at Plugin._url

pconesa: Changes were accepted here: https://github.com/I2PC/scipion-em-xmipp/pull/202

pconesa: Sorry, I'm now realizing. it was a bug. I forgot the import at #202. Thanks @dmaluenda  I misread this.

dmaluenda: Ok.

Then, if the approaching for installing "binaries" looks fine. I will merge this.

Thanks!

PR #229, opened 2020-05-08
preprocess volumes for tomography

EstrellaFG:
Generalization of preprocess protocol output in order to adapt it to tomography objects.

### DIFF
--- a/xmipp3/protocols/protocol_preprocess/protocol_process.py
+++ b/xmipp3/protocols/protocol_preprocess/protocol_process.py
@@ -162,18 +162,9 @@ class XmippProcessVolumes(ProtPreprocessVolumes):
             volumes = volInput.create(self._getPath())
             volumes.copyInfo(volInput)
             self._preprocessOutput(volumes)
-            inputVolumes = self.inputVolumes.get()
-            numberOfVols = inputVolumes.getSize()
-            for i in range(1, numberOfVols + 1):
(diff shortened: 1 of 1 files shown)


xmipp3/protocols/protocol_preprocess/protocol_process.py
```diff
@@ -159,13 +159,21 @@ def createOutputStep(self):
...
-            volumes = self._createSetOfVolumes()
+            volumes = volInput.create(self._getPath())
             volumes.copyInfo(volInput)
             self._preprocessOutput(volumes)
-            numberOfVols = self.inputVolumes.get().getSize()
+            inputVolumes = self.inputVolumes.get()
+            numberOfVols = inputVolumes.getSize()
             for i in range(1, numberOfVols + 1):
-                vol = Volume()
+                vol = inputVolumes.ITEM_TYPE()
```
pconesa, line 168: Can we use the clone method here. If so it will be more generic and Dynamic for future "volume" types.

pconesa: vol.copyAttributes

EstrellaFG: solved in https://github.com/I2PC/scipion-em-xmipp/pull/229/commits/f1a0b36b0bc1e7ac751a2421fa2ef77076657d1c

PR #230, opened 2020-05-14
Allowing multithreading in FlexiAlign

dmaluenda:
Now steps are paralyzed and multi-GPU calculation is allowed.

In addition, the validate asserts that the threads/MPIs are compatible with the number of GPUs.

I checked for threading and it works pretty well but I couldn't test it for MPI. It's returning an error but I thing it's related with my MPI configuration.

### DIFF
--- a/xmipp3/protocols/protocol_movie_correlation.py
+++ b/xmipp3/protocols/protocol_movie_correlation.py
@@ -61,6 +61,10 @@ class XmippProtMovieCorr(ProtAlignMovies):
     _label = 'correlation alignment'
     _lastUpdateVersion = VERSION_1_1
 
+    def __init__(self, **args):
+        ProtAlignMovies.__init__(self, **args)
+        self.stepsExecutionMode = cons.STEPS_PARALLEL
+
     #--------------------------- DEFINE param functions ------------------------
 
     def _defineAlignmentParams(self, form):
--- a/xmipp3/protocols/protocol_movie_correlation.py
+++ b/xmipp3/protocols/protocol_movie_correlation.py
@@ -363,13 +363,20 @@ class XmippProtMovieCorr(ProtAlignMovies):
             elif not self.useGpu.get():
                 errors.append("GPU is needed to do local alignment.")
                 return errors
-            if self.useGpu and len(self.gpuList.get().split()) +1 != self.numberOfMpi.get() * self.numberOfThreads.get():
-                coreStr = ('threads' if self.numberOfThreads.get() > 1 
-                                else ('MPI' if self.numberOfMpi.get() > 1 
-                                         else 'threads or MPI'))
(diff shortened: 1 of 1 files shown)


DStrelak:
Hi @dmaluenda ,
FlexAlign does not support parallel processing.
What we can do is multiple instantiation, using distinct GPUs. 
If that's what you want to achieve, then I would preferably remove both thread/MPI fields from the protocol, and use GPU list only (i.e. user enters N GPUs, so the workload will be divided into N batches, each batch executed on a specific GPU)

dmaluenda:
FlexAlign is not supporting paralel processing, but scipions does the work for us.

When STEP_PARALLEL active, and the number of threads is equal to gpu in use plus one, Scipion launch each step with a different device number passed in the form box. At the end, you have several calls to flexAlign binary passing a different --device number for each.

Please, test it. I did and it works.

dmaluenda:
I mean, at the end, it process each movie separatelly for the others and every one with one or another gpu number.

In other words, if you have 4 GPUs you can process 4 movies at once with 4 different calls, each in one GPU.

DStrelak:
Should we then 'bother' users with that? If we allow running on N GPUs, why does he has to set another field to N+1?
Can we just disable those (threads / MPI) fields? Then if they set N (>1) GPUs, protocol internally set threads / mpi to proper value.

cossorzano:
I agree with this

dmaluenda:
Yes and No... Yes for gpu runs, but what about the cpu implementation. For CPU, the binary below is supporting multi threading... then user can put there any number of core to use.

This is not perfect but relays a bit in Scipion more than on us... now, the validation helps users at least... if you prefer we can readjust the number of threads if GPU is chosen but it might be quite disappointing to a user putting some value and changing it.

DStrelak:
> Yes for gpu runs, but what about the cpu implementation

Hi David, 
FlexAlign (CPU) does not support threads (though it would not be that complicated to add it).
Isn't there a toggle to (not) use GPU? 
Ideally, we want sth like this:

GPU is active -> both threads and MPI is inactive
CPU is active -> MPI is active (and if user inputs 5, 5 processes will be created for movie alignment, i.e.  if we need 6 because yes, we should hand that internally).
One way or another, 'threads' field is useless.

Truth to be told, I don't know if Scipion supports that (but it definitely should).

dmaluenda:
Yes, you are right, Scipion doesn't support this dynamic active MPI section depending on GPU usage...

But, again, threads works for CPU also via Step parallelization. If user put N in threads and CPU is active, N-1 movies will be aligned at once.

And that N-1, regarding one for the master, is the common usage in Scipion. User tell us how many CPUs wants to allocate for this protocol, then if protocol needs a master, the real work will be for N-1 but this is typical.. (at least for Scipion)

dmaluenda:
I will merge this

PR #232, opened 2020-05-19
Refactor in conda manager

dmaluenda:
Simplifying the conda usage according to https://github.com/I2PC/xmipp/pull/276.

Two main things here:

1. We launch conda jobs just by launching common jobs with a proper environ.
2. We check the deepLearningToolkit is a simpler way.

### DIFF
--- a/xmipp3/base.py
+++ b/xmipp3/base.py
@@ -148,13 +148,10 @@ class XmippProtocol:
 
         # Trying to import keras to assert if DeepLearningToolkit works fine.
         kerasError = False
-        try:
-            # subprocess.call('which python', shell=True, env=env)
-            subprocess.check_output('python -c "import tensorflow"', shell=True,
-                                    env=self.getCondaEnv())
-        except subprocess.CalledProcessError as e:
-            errors.append("*Keras/Tensorflow not found*. "
-                          "Required to run this protocol.")
+        envName = CondaEnvManager.getCondaName(self, **kwargs)
+        if not os.path.isfile(getYmlTarget(envName)):
+            errors.append("*%s environment not found*. "
+                          "Required to run this protocol." % envName)
             kerasError = True
 
         # Asserting if the model exists only if the software is well installed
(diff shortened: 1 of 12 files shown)


xmipp3/base.py
```diff
@@ -148,18 +148,13 @@ def validateDLtoolkit(self, errors=None, **kwargs):
...
         # Trying to import keras to assert if DeepLearningToolkit works fine.
         kerasError = False
-        # condaEnvName preference: kwargs > protocol default > general default
-        condaEnvName = kwargs.get('_conda_env', getattr(self, "_conda_env",
-                                  CondaEnvManager.CONDA_DEFAULT_ENVIRON))
-        env = CondaEnvManager.modifyEnvToUseConda(xmipp3.Plugin.getEnviron(),
-                                                  condaEnvName)
         try:
             # subprocess.call('which python', shell=True, env=env)
-            subprocess.check_output('python -W ignore::FutureWarning '
```
rsanchezgarc, line 158: Why did you removed the -W ignore::FutureWarnings? If you remove it, they will appear messages such as feature X will be deprecated, that are not of interests for the user

dmaluenda: Now is always removed in the environ construction.

https://github.com/I2PC/xmipp/blob/8296fb03c09a61da5dea8d7d07fde027c715b974/src/xmipp/bindings/python/xmipp_base.py#L242

dmaluenda: Regarding which module we should import to check the correct installation, we can choose tensorflow or any other. It's just a quick check. Even though, we could check if the target (yml) is in the deepLearningToolkit directory, even if the environ exist. I mean, one of this checks is enough.

We shouldn't be crazy on this ;-p

rsanchezgarc: Ok. 
 I am fine with it. The problem is that Micrograph Cleaner does not use keras, only tensorflow and as we are checking for keras, the validation fails unless you install keras together with micrographCleaner.

dmaluenda:
I will merge this

PR #235, opened 2020-05-22
Metaprotocol not need protMonitor anymore

dmaluenda:
protMonitor will be moved from pwem and I've realized that mateprotocols don't need protMonitor to run, only with EMProtocol is enough.

Scipion team (@pconesa) should consider if having a metaProtocol base class is a good idea, or not. I think yes.

That base class might enclose:
  - setAborted() overrided method 
  - setFailed()  overrided method
  - _updateProtocol() useful method
  - [maybe] insertAllSteps() defining classes list and steps to do.

Well, this ideas above are for future, this Pull Request is much simpler than that.

Please, @ajimoreno check this two protocols still working in this way.

### DIFF
--- a/xmipp3/protocols/protocol_metaprotocol_discrete_heterogeneity_scheduler.py
+++ b/xmipp3/protocols/protocol_metaprotocol_discrete_heterogeneity_scheduler.py
@@ -35,7 +35,7 @@ from pyworkflow.project import Manager
 from pyworkflow.protocol import getProtocolFromDb
 import pyworkflow.object as pwobj
 
-from pwem.protocols import ProtMonitor
+from pwem.protocols import EMProtocol
 from pwem.objects import Volume
 
 from xmipp3.protocols import XmippProtSplitVolumeHierarchical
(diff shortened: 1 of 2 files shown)


pconesa:
@dmaluenda , look like there is a pattern there....What about, keeping it at xmipp plugin for now, but in the future...make an "abduction" to pwem, or even pyworkflow.

delarosatrevin:
What is the MetaProtocol? A protocol running (running internally or just controlling) other protocols? I think we had some ideas about that in the past but were not completely implemented. Anyway, any change in either pyworkflow or scipion-em should be discussed, in other to make the minimum changes and make them general enough and known by everybody.

pconesa:
Yes, @delarosatrevin we call metaportocol a protocol that creates other protocols. 2D streamer could fit into this group. My suggestion, obviously carried implicitly the agreement you mention. That why I think for now it could live at xmipp and later on, if agreed, move it somewhere else to share it with other plugins.

delarosatrevin:
Yes, that seems like a good idea. Not all creation of other protocols is
done in the same way or with the same purpose. So, we need to check if such
metaprotocol base class will provide useful functionality.

dmaluenda:
I agree with you two. In fact, even Xmipp is not needing that base class since the two meta-protocols existing in Xmipp implements all their own methods (indeed, this PR pulls down that protocols hierarchy till EMprotocol). I was just pointing a future potential idea.

Well, I have an idea of making a certain meta-protocol then, if I see the chance to make (inside Xmipp) that base collecting some meta-functionalities, I will.

pconesa:
You have a very crowded PYTHONPATH, not sure if that will help.

Do other Xmipp stuff works?

ajimoreno:
So, is it working in your computers? 
Just to know if the problem is mine or it is more general.

dmaluenda:
I think the problem is from your PYTHONPATH mess up... the log is showing several external sys.paths like prody...

I will merge this and we (@ajimoreno ) can review your configuration in a call.

PR #256, opened 2020-06-12
Co fast continuous2

cossorzano:

PR #258, opened 2020-06-16
Co plot heatmap

cossorzano:


dmaluenda:
I see similar changes in this PR and in #256, which one is the good one?

dmaluenda:
Ok, I see. This contains the #256

PR #261, opened 2020-06-19
Improving Xmipp installation

dmaluenda:
To be in sync with https://github.com/I2PC/xmipp/pull/287

@pconesa, I think it is compatible with a XMIPP_CUDA_LIB approach. This is thought to be automatic, but user could add it manually.


xmipp3/__init__.py
```diff
@@ -73,7 +78,7 @@ def getEnviron(cls, xmippFirst=True):
             environ.update({
                 'PATH': os.environ.get('CUDA_BIN', ''),
                 'LD_LIBRARY_PATH': os.environ.get('NVCC_LIBDIR', '')
-            }, position=pos)
+            }, position=pwutils.Environ.END)
```
pconesa, line 81: In another PR I'm adding CUDA_BIN and NVCC_LIBDIR (CUDA_LIB) always to the variables.

dmaluenda:
> Actually, I like this approach better. It should be more stable, in case we, for some reason, decide to rename the programs etc.

mmm.. yes. I find pros and conts in both. The cont for the other is what you say but the pro is you can check the program to use in that protocol. If the program is ready, the flag doesn't matter. For that reason I choose that one by Coss, instead this here.

PR #271, opened 2020-07-01
isXmippCudaPresent added

cossorzano:


DStrelak:
This 'clashes' with changes from @dmaluenda (see https://github.com/I2PC/scipion-em-xmipp/pull/261/files#diff-af450363ec42e8c0ff1bee73ae6f80bfR110)

dmaluenda:
Not, it doesn't clash with that. It is an alternative of my approach, but I prefer the approach in this PR. Thus, I reverted my changes and I approve this.

The implementation on the protocols should be done in a separated PR

PR #274, opened 2020-07-02
Fix monodirviewer --> _createAngDist2D, which didn't work in scipion3 and other viewer fixes

JorMaister:


pconesa:
I think this branch is no longer the right one..... @dmaluenda should this go to release branch?

DStrelak:
The target branch should be at least `devel`. `release` if you want to include it in the release :)

xmipp3/viewers/viewer_resolution_directional.py
```diff
@@ -287,7 +287,7 @@ def _createPlot(self, title, xTitle, yTitle, md, mdLabelX, mdLabelY, color='g',
         return xplotter
         
     def _showOriginalVolumeSlices(self, param=None):
-        if self.protocol.halfVolumes.get() is True:
+        if self.protocol.hasAttribute('halfVolumes'):
```
DStrelak, line 290: Generally speaking, I'm not a big fan of this approach (using strings to 'link' parts of the code), as I find it very error prone during refactoring and maintenance.

pconesa: Well, reflection is all over the place in Scipion....I'd say python nature too.

DStrelak: Yeah. Maybe we should go for more pythonic way of ['ask for forgiveness'](https://stackoverflow.com/a/610923/5484355), but that's not up to me to decide

pconesa: Interesting, although not all cases are as clear as the example....it will improve refactorization. maybe less readable in some cases.

JorMaister:
I'm not authorized to merge this branch, who should do it?

pconesa:
Leave this to xmipp guys: @dmaluenda , @DStrelak or @Vilax

PR #285, opened 2020-07-20
devel << release-20.07

dmaluenda:


dmaluenda:
There is a conflict in monores viewer.

```
<<<<<<< release-20.07
        if self.protocol.halfVolumes.get() is True:
=======
        if self.protocol.hasAttribute('halfVolumes'):
>>>>>>> devel
```

Which one is the good one, @Vilax ?

dmaluenda:
I'm not capable to resolve the conflict here.... I need help @Vilax @rmarabini (even @JorMaister ) maybe it was fixed in previous PR..

rmarabini:
Hi @dmaluenda,

  In all cases branch devel is right and branch release-20.07 is wrong. The code for visualization is now centralized in "mapVolsWithColorkey" instead of repeated in each viewer.

dmaluenda:
Thank you very much @rmarabini !!

DStrelak:
@Vilax , can you have a look at the conflict (xmipp3/viewers/viewer_resolution_monogenic_signal.py)? Thanks.

dmaluenda:
Afeter fixing the conflict, merge together with https://github.com/I2PC/xmipp/pull/297, https://github.com/I2PC/xmippViz/pull/40 and https://github.com/I2PC/xmippCore/pull/77

pconesa:
I think we need an intermediate branch, there is no way to solve conflicts in release, unless we accept all release changes.

PR #286, opened 2020-07-20
Add noise to the projection coordinates

cossorzano:


DStrelak:
@cossorzano can we merge this?

PR #288, opened 2020-07-21
Incorrect Python syntax corrected

cossorzano:


DStrelak:
This should already be fixed

PR #292, opened 2020-08-03
Outputs particles in mrcs instead of mrc. So dimensions are right. TO…

pconesa:
…DO: Scipion should be clever to Know that these are 2D objects.


dmaluenda:
Great, at the end, just replacing mrc for mrcs does the magic. Awsome!

DStrelak:
Please, no _TODO_ in the PR :-D

pconesa:
It was a TODO for scipion.

PR #295, opened 2020-08-07
Fix some tests

ajimoreno:
Some bugs fixed


pconesa:
Thanks @ajimoreno let's see what buildbot says.

PR #296, opened 2020-08-10
Cleaning some intermediate files

cossorzano:


dmaluenda:
against devel is right? Isn't it a hot fix for the release?

PR #306, opened 2020-09-08
Fixing an error with the dictionary keys

fonsecareyna82:
Fixing an error with the dictionary keys

### DIFF
--- a/xmipp3/protocols/protocol_align_volume.py
+++ b/xmipp3/protocols/protocol_align_volume.py
@@ -215,11 +215,11 @@ class XmippProtAlignVolume(ProtAlignVolume):
             outputArgs = {'outputVolumes': volSet}
         else:
             vols[0].setSamplingRate(self.inputReference.get().getSamplingRate())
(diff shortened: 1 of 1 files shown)


DStrelak (approved):

dmaluenda:
I will cancel this PR because some tests and templates expect outputVolume as the current output.

Let's do it in devel in order to have time to accommodate the rest of stuff.

xmipp3/protocols/protocol_align_volume.py
```diff
@@ -215,11 +215,11 @@ def createOutputStep(self):
             outputArgs = {'outputVolumes': volSet}
```
dmaluenda, line 215: I would change this for
```
outputArgs = {'outputVolume': volSet}
```

In this way, the option that was working still same

fonsecareyna82: You have reason. I will change !!!

dmaluenda (approved):

PR #309, opened 2020-09-16
mm_ChimeraX

mmmtnez:
Hi all,

small modifications in the extract asymmetric unit viewer to adapt it to ChimeraX


rmarabini (approved):

rmarabini:
I made a mall modification so innerRadius is set to 0 is negative. Otherwise the color map does not go from 0 to manRadius as should be.

PR #310, opened 2020-09-18
show unit cell does not work 

rmarabini:
Show unit cell does not work. It does not work when  minradius=-1, which is the default value.

 This is partially fixed in @mmmtnez  PR but I like better my  solution since the value of minradius is used in some calculations and should not be -1 but 0


rmarabini:
I think is less confusing if a mody mmmtnz PR

PR #314, opened 2020-09-30
fixing automatic control points

DStrelak:


JorMaister (approved):

dmaluenda:
Is this related with the performance issue? If so, great! Well done guys!

dmaluenda (approved):

DStrelak:
> Is this related with the performance issue? If so, great! Well done guys!

Yes... indirectly. The number of control points was badly calculated, coz I can't tell the difference between multiplication and division. The underlying problem, that the equation solver is way too slow, is still on the TODO list.

PR #316, opened 2020-10-01
Rsg deep e mhancer

rsanchezgarc:


dmaluenda:
This PR contains commits from devel. To make a Pr only with your commit, run
```
git checkout release-20.07
git pull
git checkout -b rsg_emEnhanserRelease  # or the name of your choice
git cherry-pick 3bb202fe63624999de386184668f9745ceb2bc06
git push  # probably it will not work but it will suggest how to push
```
and make the PR from that branch.

I'm closing these two PRs. This will be replaced with the one described here and the other due to that commit will arrive to devel in a couple of days when we merge release to devel.

Thanks!

PR #321, opened 2020-10-09
Mm asymmetric unit

mmmtnez:
Hi all,

an additional small change in the extract asymmetric unit protocol that still had the old name in labels.


mmmtnez:
I did not realize that the merging branch was devel. sorry

PR #322, opened 2020-10-09
Mm asymetric unit

mmmtnez:
Hi all,

an additional small change to rename labels of extract asymmetric unit that still had the old name.


rmarabini:
This PR is confusing. Mainly is identical to a previous PR. I am going to delete it and recreate. Most of the changes will be gone.

PR #324, opened 2020-10-14
devel << release-20.07

dmaluenda:
Conflicts can be fixed directly in release-20.07 since its code is released in the master branch.


dmaluenda:
Conflicts fixed

PR #327, opened 2020-11-04
Aj monores half volumes

ajimoreno:
I have modified MonoRes protocol to allow the option of using the half maps stored in the input volume. I have tried the different options and everything seems to work fine, but take a careful look at it, because I'm not familiar with the details of this protocol.


cossorzano (dismissed):
OK

dmaluenda (approved):

cossorzano (approved):

PR #329, opened 2020-11-05
Removed pyvista dependency and associated methods

DavidHerreros:
Requirements and utils have been modified to remove the dependency introduced in previous PR


pconesa (approved):

pconesa:
this PR is to remove pyvista dependency from xmipp plugin. This code will go to a new tomo plugin @DavidHerreros is preparing.

PR #333, opened 2020-11-11
Fix fscq blocres

erneyramirez:
changes in fsc-q protocol to fix possible incompatibility with bsoft versions.


dmaluenda (approved):
I agree on that changes.

I'm waiting just in case we want to include some minor regarding the cuda-support PR on xmipp repository.

PR #336, opened 2020-11-30
Ds fixing flex align protocol issues

DStrelak:


pconesa (approved):
It looks correct, at least for the "spaces" issue.

The only thing I can say is that that part (the one fixing the spaces in the path) could be a hotfix and be released and fixed today. Probably not the rest fo the code.

cossorzano (approved):

PR #340, opened 2020-12-17
use an "indirect" pointer to mics in ctf set. This should fix pointer…

pconesa:
… being "moved" to the wrong attribute.

Also, consider different size of ctf sets as input (double consensus where first has discarded some ctfs)

Hi, this should fix 2 couple of issues. Is up to you if you want this sent to master or devel.

I started from master since to me it is a "hotfix"


DStrelak (approved):

DStrelak:
I agree  both with changes and fixing release.  However, I would wait till https://github.com/I2PC/scipion-em-xmipp/issues/342 is fixed as well, as it's a required input for ctf consensus.

cossorzano (approved):

Issue #338, opened 2020-12-02
deepPostProcessing fails to create output starting from two half maps


mmmtnez:
Detailed below appears the error in protocol_postProcessing_deepPostProcessing in createOutputStep
 self._defineTransformRelation(self.inputHalf1, volume)

Neural net inference
01259:   WARNING: This is xmipp_base implementation for script
01260:   deepemhancer  -i Runs/008142_XmippProtDeepVolPostProc/tmp/inputHalf1.mrc  -i2 Runs/008142_XmippProtDeepVolPostProc/tmp/inputHalf2.mrc  -o Runs/008142_XmippProtDeepVolPostProc/extra/deepPostProcess.mrc  --deepLearningModelPath /home/roberto/scipion3/software/em/xmipp/models/deepEMhancer/production_checkpoints/deepEMhancer_tightTarget.hd5  --samplingRate 1.430000 --cleaningStrengh -1.000000 -g 0 -b 8 
01261:   deepemhancer  -i Runs/008142_XmippProtDeepVolPostProc/tmp/inputHalf1.mrc  -i2 Runs/008142_XmippProtDeepVolPostProc/tmp/inputHalf2.mrc  -o Runs/008142_XmippProtDeepVolPostProc/extra/deepPostProcess.mrc  --deepLearningModelPath /home/roberto/scipion3/software/em/xmipp/models/deepEMhancer/production_checkpoints/deepEMhancer_tightTarget.hd5  --samplingRate 1.430000 --cleaningStrengh -1.000000 -g 0 -b 8 
01262:   FINISHED: deepVolPostProStep, step 2, time 2020-12-02 14:21:12.612273

ERROR:   

01263:   STARTED: createOutputStep, step 3, time 2020-12-02 14:21:13.122197
01264:   Traceback (most recent call last):
01265:     File "/home/roberto/scipion3/.scipion3/lib/python3.7/site-packages/pyworkflow/protocol/protocol.py", line 187, in run
01266:       self._run()
01267:     File "/home/roberto/scipion3/.scipion3/lib/python3.7/site-packages/pyworkflow/protocol/protocol.py", line 238, in _run
01268:       resultFiles = self._runFunc()
01269:     File "/home/roberto/scipion3/.scipion3/lib/python3.7/site-packages/pyworkflow/protocol/protocol.py", line 234, in _runFunc
01270:       return self._func(*self._args)
[...]

PR #344, opened 2020-12-22
bug fixing and tests issue #338

rsanchezgarc:
Fixed bug in input from half-maps
New test for half-maps as input


cossorzano:
OK

cossorzano (approved):

PR #349, opened 2021-01-12
2 scipion3 mm localdeblur

mmmtnez:
Hi all,

here you have some small changes in localdeblur protocol to avoid copying the last iteration map and to add the origin to all iteration maps, not only to the last one. Now the last iteration will not appear twice in the extra folder, once with the number of iteration as suffix and once with the "last" suffix. It will only appear with the "last" suffix. Some users did not understand the results obtained and, moreover, they found difficult to select the last iteration as input for the following protocols, since it did not appear as "last" in the output. An additional test has been created to check localdeblur results in case more than one iteration map could be generated.

Changes from tmp to extra folder have been added to allow the visualization of results in ChimeraX (cxc file now will be saved in the extra folder).

Test: 

scipion3 tests xmipp3.tests.test_protocols_localdeblur.TestLocalDeblur


erneyramirez:
Por qué son todos estos cambios en LocalDeblur???

rmarabini:
> Por qué son todos estos cambios en LocalDeblur???
> Siempre he pensado que en el mapa, los ejes del origen le quitan visibilidad, lo mismo pienso para la resolución y por eso no se la puse a deepRes. Yo creo que si el usuario quiere los ejes porque los necesita, puede ponerlos él. Creo que ahora eso es bastante fácil en chimeraX.

Hi Erney,

   I do not understand this comment since no change is introduced in the viewer. Mainly there are two changes. The first one copies the 3D input map origin to the output 3D maps. The second change avoid the creation of an extra copy of the last output volume since it is not needed. Neither of these two modification  change how the volume is visualized and the second one save a bit of disk space. Note that if we the volume origin is not copied from the input volume to the output then it is much more difficult to use this sharpening  for model building.

rmarabini (approved):

erneyramirez:
Ok! I thought you were referring to the display of the origin axes.

rmarabini:
> Ok! I thought you were referring to the display of the origin axes.

Can we merge the changes?

PR #350, opened 2021-01-12
change tmp by extra dir in viewer using chimera

EstrellaFG:


cossorzano:
Estrella, which is the advantage of generating the visualization files in extra and not tmp? They occupy space.

EstrellaFG:
> Estrella, which is the advantage of generating the visualization files in extra and not tmp? They occupy space.

It is because tmp now disapear after execution, thus, @pconesa indicated us to change the files that viewers will use (ChimeraX) from tmp to extra.

pconesa:
Yes, tmp folder is not longer available after the protocol finishes (a consequence of the SSD feature).

So, viewers are using extra. They could use "root" path if better. Or even they could create the tmp folder again if needed. In any case, they occupy the same space in extra or in tmp.

cossorzano:
The advantage of using tmp was that I presumed that from time to time, 
Scipion was deleting all files found in tmp directories. Was that 
correct? If so, they could be written in a project tmp directory. Would 
that be an option?

Cheers, Carlos Oscar

pconesa:
Scipion does not clean tmp folder AFTER execution is finished. I don't think project/tmp is cleaned either.

If you want a temporary, why not using standard python tmp folder, that goes to system tmp --> /tmp?

using: tempfile.mkdtemp
or just tempfile.gettempdir() if you want to work directly on /tmp.

https://docs.python.org/3/library/tempfile.html

EstrellaFG:
I can try that approach, thanks!

PR #352, opened 2021-01-13
Change extra dir by  standard python tmp folder in viewer

EstrellaFG:


pconesa (commented):
I'd take the chance to improve a bit (easy stuff) the code.

join(tempfile.gettempdir(), 'tmpVolumes.sqlite') --> do it only once, at the beginning and then use a variable.

I'd also go for less generic names, instead of "chimera.cxc" something like "vol_subtact_chimera.cxc" to avoid name collisions.

I'd say one more thing.... is this process "time consuming". Do we want to generate those files over and over everytime viewer is launched?

xmipp3/viewers/viewer_volume_subtraction.py
```diff
@@ -81,23 +82,23 @@ def _showVolumes(self, paramName=None):
             return self._showVolumesXmipp()
 
     def _createSetOfVolumes(self):
-        if not exists(self.protocol._getExtraPath('tmpVolumes.sqlite')):
-            tmpFileName = self.protocol._getExtraPath("tmpVolumes.sqlite")
+        if not exists(join(tempfile.gettempdir(), 'tmpVolumes.sqlite')):
```
pconesa, line 85: Actually I do not think you want to create a set of volume in tmp if they exist in extra. This is wrong

pconesa: Oh, wait, is this like "avoiding" the creation of the set if it exists?

pconesa: Forget this, my bad.

pconesa (dismissed):

pconesa (approved):

xmipp3/viewers/viewer_volume_subtraction.py
```diff
@@ -83,13 +83,11 @@ def _showVolumes(self, paramName=None):
...
     def _createSetOfVolumes(self):
         tmpFileName = join(tempfile.gettempdir(), 'tmpVolumes_adjust.sqlite')
-        if not exists(tmpFileName):
-            _outputVol = self.protocol.outputVolume
-            setOfVolumes = SetOfVolumes(filename=tmpFileName)
-            setOfVolumes.append(_outputVol)
-            setOfVolumes.write()
-        else:
-            setOfVolumes = SetOfVolumes(filename=tmpFileName)
+        os.remove(tmpFileName)
```
DStrelak, line 86: Are you sure that deleting the directory is the right step? Do you know that there are no additional data?

pconesa: tmpFileName is not a directory, is the sqlite file, which very likely would have been generated by this same code before.

EstrellaFG: Yes, I was checking that an tmpFileName is a file (sqlite). I am doing that because I have observed that sometimes occurs that the tmp directory was not cleaned before I run the viewer again and thus, the viewer show the result of other execution (other "box" of this protocol of the same or different project that has been executed before).

DStrelak: Ah, my bad

cossorzano (approved):

Issue #347, opened 2021-01-05
XmippProjMatchViewer object has no attribute createVolumesSqlite


fonsecareyna82:
XmippProjMatchViewer object has no attribute createVolumesSqlite

PR #365, opened 2021-03-01
Fix viewers

DaniDelHoyo:
Viewers in scipion-em-xmipp corrected for using the extra path instead of the tmp path.


pconesa:
Closes #347

dmaluenda (dismissed):
All make sense

dmaluenda (approved):

cossorzano (commented):
OK

PR #379, opened 2021-03-15
deep align protocol

ajimoreno:
The protocol for deepAlign is included here


xmipp3/wizards.py
```diff
@@ -29,7 +29,7 @@
 from pwem.constants import *
 from pwem.wizards import *
 from pyworkflow.wizard import Wizard
-from xmipp3.viewers import XmippMonoResViewer, XmippMonoTomoViewer, XmippResDeepResViewer
```
DStrelak, line 32: this looks suspicious

DStrelak (approved):

PR #380, opened 2021-03-18
Movie gain tiff flipping

DaniDelHoyo:
Manages the tiff movies flipping issues.

### DIFF
--- a/xmipp3/protocols/protocol_movie_correlation.py
+++ b/xmipp3/protocols/protocol_movie_correlation.py
@@ -45,7 +45,7 @@ from pwem.protocols import ProtAlignMovies
 
 from xmipp3.convert import writeMovieMd
 from xmipp3.base import isXmippCudaPresent
-from ..utils import *
+import xmipp3.utils as xmutils
 
 
 class XmippProtMovieCorr(ProtAlignMovies):
@@ -322,7 +322,7 @@ class XmippProtMovieCorr(ProtAlignMovies):
         baseName = os.path.basename(gainFn).replace(ext, '_transformed' + ext)
         outFn = os.path.abspath(self._getExtraPath(baseName))
 
-      gainImg = readImage(gainFn)
+      gainImg = xmutils.readImage(gainFn)
       imag_array = np.asarray(gainImg.getData(), dtype=np.float64)
 
       # Building the transformation matrix
@@ -330,8 +330,8 @@ class XmippProtMovieCorr(ProtAlignMovies):
       M = self.getUserFlip(imag_array)
       print('Transforming gain: {} degrees rotation, {} flip'.
             format(angle, ['no', 'vertical', 'horizontal'][self.gainFlip.get()]))
-      flipped_array, M = rotation(imag_array, angle, imag_array.shape, M)
-      writeImageFromArray(flipped_array, outFn)
+      flipped_array, M = xmutils.rotation(imag_array, angle, imag_array.shape, M)
+      xmutils.writeImageFromArray(flipped_array, outFn)
       return outFn
 
     def flipYGain(self, gainFn, outFn=None):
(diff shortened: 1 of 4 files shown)


DaniDelHoyo:
As a minor comment, the functions flipYGain, applyTransform, rotation, readImage are repeated and should be placed in some utils file and imported.

DaniDelHoyo:
As a minor comment, the functions flipYGain, applyTransform, rotation, readImage are repeated and should be placed in some utils file and imported. Which "utils" file do you think we can send them to?

xmipp3/protocols/protocol_movie_correlation.py
```diff
@@ -408,4 +432,40 @@ def _validate(self):
...
+def applyTransform(imag_array, M, shape):
+  ''' Apply a transformation(M) to a np array(imag) and return it in a given shape
+  '''
+  imag = emlib.Image()
+  imag.setData(imag_array)
+  imag = imag.applyWarpAffine(list(M.flatten()), shape, True)
+  return imag.getData()
+
+
+def rotation(imag, angle, shape, P):
```
DStrelak, line 458: this protocol is not a right place for this code. Here it cannot be reused

xmipp3/protocols/protocol_movie_correlation.py
```diff
@@ -408,4 +432,40 @@ def _validate(self):
...
+        img.setData(array)
+        img.write(fn)
+
+    def readImage(self, fn):
+        img = emlib.Image()
+        img.read(fn)
+        return img
+
+
+def applyTransform(imag_array, M, shape):
```
DStrelak, line 449: doesn't seem to be called

xmipp3/protocols/protocol_movie_gain.py
```diff
@@ -453,6 +454,21 @@ def match_orientation(self, exp_gain, est_gain):
             #self.writeImageFromArray(best_gain_array, self.getBestCorrectionPath())
 
     # ------------------------- UTILS functions --------------------------------
+    def flipYGain(self, gainFn, outFn=None):
```
DStrelak, line 457: duplicated code

xmipp3/protocols/protocol_movie_correlation.py
```diff
@@ -249,7 +267,13 @@ def tryProcessMovie(self, movie):
             args += ' --dark "%s"' % self.inputMovies.get().getDark()
 
         if self.inputMovies.get().getGain():
-            args += ' --gain "%s"' % self.inputMovies.get().getGain()
+            ext = pwutils.getExt(self.inputMovies.get().getFirstItem().getFileName()).lower()
+            if ext in ['.tif', '.tiff']:
+              self.flipY = True
```
DStrelak, line 272: what if I don't want to flip the gain? This should be settable from GUI

DStrelak (changes requested):

xmipp3/protocols/protocol_movie_gain.py
DaniDelHoyo: yes, as I have said in the comments, this functions should be moved into some "utils" file so they are not duplicated and can be reused, but I don't know where exactly. Any suggestions?

xmipp3/protocols/protocol_movie_correlation.py
DaniDelHoyo: Indeed I think it would be a good idea to set an extra GUI option for additional flips or rotations as there are in other movie alignment protocols such as motioncorr2, unblur, relion. 
However, we already had a big discussion ( https://github.com/scipion-em/scipion-em/pull/202 ) about how to manage the flips and it was decided to address it this way for tiff movies, assuming you have used movie gain it will work. Assuming you have not used movie gain, flipping it or not will not guaranty a good gain correction.

DStrelak:
> Which "utils" file do you think we can send them to?

No idea, the best candidate is `convert/convert.py`, but I don't think it's the right place. Try asking somebody from Scipion team, this looks like a sufficiently general functionality.

xmipp3/protocols/protocol_movie_correlation.py
DStrelak: TLDR (scipion-em/scipion-em#202 );
If we do something by default for user, it makes user happy. If user cannot change default value at all, user is miserable. 
It has to be possible to turn off automatic flip from GUI.

DaniDelHoyo: I agree user must be able to perform further rotations or flips as in other movie alignment protocols. I will implement them in the next version, while I relocate the duplicated functions

DaniDelHoyo: user happy now

xmipp3/protocols/protocol_movie_gain.py
DStrelak: so why don't you extract this `flipYGain` into the `xmipp3/utils.py` as e.g. `flip(file, newName)`?

DStrelak (approved):
I've also updated sonarcloud rules, you should see less code smells after next analysis

DaniDelHoyo:
The test of this protocol has been adapted and the flipYGain (now flipYImage) has been moved to xmipp3/utils.
If now further changes are requested I think we can successfully and finally approve this PR

DStrelak (approved):

PR #389, opened 2021-04-14
Fix tests

EstrellaFG:
- Remove monotmo test and wizard (monotomo protocol is in xmipptomo plugin)
- Fix parameter name in protocol pdbconvert


xmipp3/tests/test_protocol_monotomo.py
```diff
@@ -1,81 +0,0 @@
-# **************************************************************************
```
DStrelak, line 1: Deleted on purpose?
@Vilax

EstrellaFG: it was in the description of the PR ;) => Remove monotmo test and wizard (monotomo protocol is in xmipptomo plugin)

DStrelak (approved):

Issue #408, opened 2021-05-04
Remove OpenCV dependency


DStrelak:
We should replace 
https://github.com/I2PC/scipion-em-xmipp/blob/devel/xmipp3/protocols/protocol_deep_align.py#L472
 by our own implementation to avoid openCV dependency.
See https://github.com/I2PC/scipion-em-xmipp/blob/devel/xmipp3/utils.py#L72

PR #409, opened 2021-05-04
Ignore viewer for protocol_deep_align.py 

DaniDelHoyo:
until cv2 dependency is removed.

### DIFF
--- a/xmipp3/protocols/protocol_deep_align.py
+++ b/xmipp3/protocols/protocol_deep_align.py
@@ -435,7 +435,6 @@ _noiseCoord   '0'
         fnLabels = self._getExtraPath('labels.txt')
         mdIn = emlib.MetaData(fnProj)
         mdExp = emlib.MetaData()
-        newImage = emlib.Image()
         maxPsi = 180
         maxShift = round(newXdim / 10)
         idx = 1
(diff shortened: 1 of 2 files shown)


DaniDelHoyo:
These last changes should remove the cv2 dependency. However, I am not 100% sure that cv2.warpAffine and our ImageHandler applyTransform will equally behave in these transformations, especially because of the borderMode argument, which is not the same as our doWarp.

cossorzano:
This difference at the borders should not make an important difference in the network training

cossorzano (approved):

xmipp3/protocols/protocol_deep_align.py
```diff
@@ -469,11 +468,14 @@ def generateExpImagesStep(self, Nimgs, nameProj, nameExp, label):
...
-                    newFn = ('%06d@' % idx) + fnExp[:-3] + 'stk'
-                    newImage.setData(newImg)
-                    newImage.write(newFn)
+                    #newImg = cv2.warpAffine(I.getData(), M, (Xdim, Ydim),
+                    #                        borderMode=cv2.BORDER_REFLECT_101)
+                    #newFn = ('%06d@' % idx) + fnExp[:-3] + 'stk'
+                    #newImage.setData(newImg)
+                    #newImage.write(newFn)
+
+                    self._ih.applyTransform(fnImg, newFn, M, (Ydim, Xdim), doWrap=True)
```
DStrelak, line 477: these changes are not equivalent. newImage is not updated. newFn is not written.

DStrelak (changes requested):

xmipp3/protocols/protocol_deep_align.py
DaniDelHoyo: newImage was just used for saving the newFn, which it is written. But I do commented the definition of newFn, it is corrected now. And the commented lines are removed.

DStrelak (approved):

DaniDelHoyo:
Because of previous "code smells" (I din't introduce any new) it cannot be directly merged. @DStrelak , could you please take a look?
It solves #https://github.com/I2PC/scipion-em-xmipp/issues/408

PR #412, opened 2021-05-24
add compute energy flag

EstrellaFG:
This protocol also uses xmipp program volume_subtraction and I need to do this modification as I modify computeEnergy flag as requested in PR https://github.com/I2PC/xmipp/pull/388


DStrelak (approved):

xmipp3/protocols/protocol_volume_adjust_sub.py
```diff
@@ -278,6 +283,8 @@ def adjustStep(self):
             args += ' --cutFreq %f --sigma %d' % (fc, self.sigma.get())
```
DStrelak, line 283: Just a thought: this `adjustStep` and second half of the `insertAllSteps` are rather similar. How about extracting a method which would invoke the volume substraction?

PR #413, opened 2021-06-03
Remove SPH Code

DavidHerreros:
Protocols, tests, and viewers associated with the Zernikes3D algorithm have been commented so they are not offered in Scipion when Xmipp is released.


DStrelak (changes requested):
Please add a comment to each file why are those lines commented out, or remove them completely (whichever makes more sense to you)

DavidHerreros:
> Please add a comment to each file why are those lines commented out, or remove them completely (whichever makes more sense to you)

I would prefer the codes to be commented out to make merges in their future in an easier way in the future. I have added a small comment at the beginning of the protocols, tests, and viewers explaining that they will be available again in a future release.

DStrelak (approved):

PR #416, opened 2021-06-21
Update protocol_create_mask3d.py

pconesa:
This will create masks in mrc format instead of .vol which os more likely to be compatible with most software.

### DIFF
--- a/xmipp3/protocols/protocol_preprocess/protocol_create_mask3d.py
+++ b/xmipp3/protocols/protocol_preprocess/protocol_create_mask3d.py
@@ -189,7 +189,7 @@ sph + 1 '3.03623188  0.02318841 -5.04130435' '7'
 
     #--------------------------- INSERT steps functions ------------------------
     def _insertAllSteps(self):
-        self.maskFile = self._getPath('mask.vol')
+        self.maskFile = self._getPath('mask.mrc')
         
         if self.source == SOURCE_VOLUME:
             self._insertFunctionStep('createMaskFromVolumeStep')


pconesa:
Hi! this is a "trivial change". But you are involved as reviewers in case you see there is an extra complication I might have missed in the case of changing masks from .vol to .mrc? in a later usage of it.

Vilax (dismissed):

rmarabini (approved):

rmarabini (dismissed):
I see no problem wit this change. Although I am not sure mrc format should be prioritized.

DStrelak:
I also don't (fore)see any problem. Though I have to say that 
```
def _insertAllSteps(self):
        self.maskFile = self._getPath('mask.mrc')
``` 
Feels more like you expect to find `mask.mrc`, instead of creating one.

Is this expected to be part of the release?

pconesa:
This was triggered by pyseg not being able to read -vol (or at least not trying to and only accepting .rec, .mrc and .st)

So, most likely, mrc will be compatible with most software than .vol

Vilax:
From my point of view, we should remove all .vol that Xmipp generates and substitute them by .mrc. I don't know if it is the best format, but for sure is the most used. I think it avoid compatibility problems or converts of many protocols.

rmarabini:
> pyseg



> This was triggered by pyseg not being able to read -vol (or at least not trying to and only accepting .rec, .mrc and .st)
> 
> So, most likely, mrc will be compatible with most software than .vol

@pconesa ,

   I do not think this is a good argument. If pyseg cannot handle the ".vol" extension then the protocol needs a convert step at the beginning, sooner or latter some user will create a mask with  another program that it is not under our control and does not produce mrc. From my point of view all protocols should provide a convert step when needed.

   On the other hand I have no problem if scipion suggest a (or a few) preferred formats. I agree with @Vilax  that MRC is the more popular although not the best format.

pconesa:
I know, and pyseg will implement a conversion in case it finds a .vol

But using the most compatible format, will prevent conventions, and errors, when the developer miss to implement the conversion step.

rmarabini:
In case there is any doubt, this PR is OK with me

Vilax (commented):
Why 21.04 instead of 21.09?

mmmtnez:
I agree with this change since the protocolo relion subtract projection did not recognize the extension .vol for the mask, it prefers .mrc also

mmmtnez (approved):

PR #421, opened 2021-06-29
FlexAlign: fixing default number of control points

DStrelak:


DStrelak:
@JorMaister this change might affect also tomo plugins, so I will not include it in this release. 
Please keep this change in mind in case you see some weird behavior.

JorMaister (approved):

cossorzano (approved):

Issue #430, opened 2021-08-19
protocol 'xmipp3 - align volume' fails when mask given


JuhaHuiskonen:
The alignment fails when a mask is given with error:

00019:       maskArgs+=" --mask binary_file %s" % self.volMask
00020:   AttributeError: 'XmippProtAlignVolume' object has no attribute 'volMask'

Scipion 3.0.8

PR #432, opened 2021-08-25
Fix protocol_align_volume binary mask

EstrellaFG:
Fix binary mask option of protocol_align_volume and add tests for circular and binary mask options that were missing.
The bug is reported in issue: https://github.com/I2PC/scipion-em-xmipp/issues/430

### DIFF
--- a/xmipp3/tests/test_protocols_xmipp_3d.py
+++ b/xmipp3/tests/test_protocols_xmipp_3d.py
@@ -48,6 +48,10 @@ from xmipp3.protocols.protocol_align_volume import (ALIGN_ALGORITHM_EXHAUSTIVE,
                                                ALIGN_ALGORITHM_EXHAUSTIVE_LOCAL,
                                                ALIGN_ALGORITHM_LOCAL)
 
+MSG_WRONG_SAMPLING = "wrong sampling rate"
+MSG_WRONG_SIZE = "wrong size"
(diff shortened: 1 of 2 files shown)


xmipp3/tests/test_protocols_xmipp_3d.py
```diff
@@ -789,6 +793,56 @@ def testExhaustiveLocal(self):
...
+                                     stepRotationalAngle=10,
+                                     minTiltAngle=65,
+                                     maxTiltAngle=100,
+                                     stepTiltAngle=10,
+                                     minInplaneAngle=0,
+                                     maxInplaneAngle=0,
+                                     stepInplaneAngle=0,
+                                     applyMask=True,
+                                     maskType=1,
+                                     numberOfMpi=1, numberOfThreads=1
```
DStrelak, line 840: will there be any execution time improvement if we set threads / mpi to higher values?

DStrelak (approved):

xmipp3/tests/test_protocols_xmipp_3d.py
EstrellaFG: It takes between 3 and 7 seconds per test, in total the 5 tests are 34 seconds, so I think it is not necessary to set higher values. Moreover, AFAIK buildbot doesn work with mpi.

PR #433, opened 2021-08-27
TestXmippDenoiseParticles: fixing the execution on Buildbot

DStrelak:
Disabling MPI. The new beta relion cannot run with MPI:
```
ERROR:
Protocol has validation errors:
Gradient refinement (running the VDAM algorithm) is not supported together with MPI
```

Since this protocol is very fast, it seems to be a reasonable solution


danielmarchan3 (approved):
I checked the files changed and run the TestXmippDenoiseParticles locally and it seems to work correctly.

PR #440, opened 2021-09-29
fix output location + cleaning

EstrellaFG:


DStrelak (approved):

DStrelak:
Is this intended for devel branch only?

EstrellaFG:
> Is this intended for devel branch only?

no, there is a PR in release branch also, already merged I think

PR #449, opened 2021-11-04
Zernike programs: replace .vol by .mrc, improving tests stability

DavidHerreros:
Zernike programs now generate output and intermediate volumes in `.mrc` format instead of `.vol`.

Tests stability of Zernike programs has also been improved.

### DIFF
--- a/xmipp3/protocols/protocol_volume_deform_zernike3d.py
+++ b/xmipp3/protocols/protocol_volume_deform_zernike3d.py
@@ -156,7 +156,7 @@ class XmippProtVolumeDeformZernike3D(ProtAnalysis3D):
 
     def convertOutputStep(self):
         fnOutVol = self._getFileName('fnOutVol')
-        params = ' -i %s --sampling_rate %.2f' % (fnOutVol, self.newTs)
+        params = ' -i %s --sampling_rate %f' % (fnOutVol, self.newTs)
         self.runJob("xmipp_image_header", params)
 
     def createOutputStep(self):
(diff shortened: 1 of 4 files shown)


xmipp3/protocols/protocol_volume_deform_zernike3d.py
```diff
@@ -153,6 +154,10 @@ def deformStep(self):
                 params = params + ' --thr %d' % self.numberOfThreads.get()
             self.runJob("xmipp_volume_deform_sph", params)
 
+    def convertOutputStep(self):
```
DStrelak, line 157: is this change here on purpose?

DStrelak (approved):

xmipp3/protocols/protocol_volume_deform_zernike3d.py
cossorzano: No that I remember

DavidHerreros: Yes, I added it to ensure that the output volume that we are generating has the correct sampling on the mrc file header based on the user's parameters.

If you consider that this is not necessary, I will remove it from the protocol.

xmipp3/protocols/protocol_volume_deform_zernike3d.py
```diff
@@ -153,6 +154,10 @@ def deformStep(self):
...
+        fnOutVol = self._getFileName('fnOutVol')
+        params = ' -i %s --sampling_rate %.2f' % (fnOutVol, self.newTs)
```
cossorzano, line 159: I would not limit the string to %.2f to avoid loss of precision, better %f

DStrelak (approved):

PR #453, opened 2021-11-10
FSC-Q: fixing circular save issue when creating output

delarosatrevin:
It seems that there is a bug in the master branch (and maybe also devel) of the 'xmipp - validate fsc-q' protocol. My guess is that all the calls to self._store after the defineOutput are causing a circular reference issue. 

```
STARTED: createOutputStep, step 6, time 2021-11-10 14:32:03.826672
Object.getAttributesToStore: attribute '_origin' seems to be overwritten,
   since '_objDoStore' was not found. Ignoring attribute.
Object.getAttributesToStore: attribute 'fscq_Volume' seems to be overwritten,
   since '_objDoStore' was not found. Ignoring attribute.
Protocol failed: Circular reference, object: 55.96._origin found twice
FAILED: createOutputStep, step 6, time 2021-11-10 14:32:04.049656
------------------- PROTOCOL FAILED (DONE 6/6)
*** Last status is failed
Object.getAttributesToStore: attribute '_origin' seems to be overwritten,
   since '_objDoStore' was not found. Ignoring attribute.
Object.getAttributesToStore: attribute 'fscq_Volume' seems to be overwritten,
   since '_objDoStore' was not found. Ignoring attribute.
```

The changes in this PR move the creation of metrics values to a separate function and store all these values (within a single db commit) in the defineOutputs function.


DStrelak:
Hi @delarosatrevin ,
Thanks for contributing!
Can I ask you to change the base branch to devel? We don't do PRs against master.

cossorzano:
Merged into devel

DStrelak:
> Merged into devel

See https://github.com/I2PC/scipion-em-xmipp/pull/468

delarosatrevin:
@DStrelak thanks for merging the changes...but in general you should consider allowing hotfixes to go quicker to master and then merge then with devel. Especially with bugs as critical as this one, that prevents any user to run the program in the master version.

pconesa:
I agree with @delarosatrevin  there are many bugfixes that can go quickly and straight away to master and be released on the same day or week . Be brave!

PR #455, opened 2021-11-23
Volume consensus: enhanced validation and error reporting

EstrellaFG:
raise exception if no volume is generated

### DIFF
--- a/xmipp3/protocols/protocol_volume_consensus.py
+++ b/xmipp3/protocols/protocol_volume_consensus.py
@@ -70,7 +70,8 @@ class XmippProtVolConsensus(ProtInitialVolume):
         outVol.setSamplingRate(self.vols[0].get().getSamplingRate())
         outVol.setFileName(self._getExtraPath("consensus_volume.mrc"))
         if not exists(self._getExtraPath("consensus_volume.mrc")):
-            raise NoOutputGenerated
+            raise NoOutputGenerated("Consensus volume NOT generated, please check input volumes to ensure they have "
+                                    "equal box size, voxel size and origin.")
         else:
             outVol2 = Volume()
             outVol2.setSamplingRate(self.vols[0].get().getSamplingRate())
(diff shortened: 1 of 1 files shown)


xmipp3/protocols/protocol_volume_consensus.py
```diff
@@ -69,11 +70,15 @@ def createOutputStep(self):
         outVol = Volume()
         outVol.setSamplingRate(self.vols[0].get().getSamplingRate())
         outVol.setFileName(self._getExtraPath("consensus_volume.mrc"))
-        outVol2 = Volume()
-        outVol2.setSamplingRate(self.vols[0].get().getSamplingRate())
-        outVol2.setFileName(self._getExtraPath("consensus_volume_diff.mrc"))
-        self._defineOutputs(outputVolume=outVol)
-        self._defineOutputs(outputVolumeDiff=outVol2)
+        if not exists(self._getExtraPath("consensus_volume.mrc")):
+            raise Exception("Consensus volume NOT generated, please check input volumes to ensure they have equal box "
```
DStrelak, line 74: Don't throw generic exception
Also, what is the reason that the `consensus` file is not produced? If some previous step fails, than the output step should not even be called, right?
Can you check the box size / voxel size and origin requirements in the `_validate(self)` method?

DStrelak (changes requested):
Please follow the naming convention of the PR

xmipp3/protocols/protocol_volume_consensus.py
EstrellaFG: The xmipp program called in runJob does not give any error, it just does not generate the output volume if something went wrong when computing the wavelet transform (with a external library). In principle, with the fix in https://github.com/I2PC/xmipp/pull/473#issuecomment-976340917 it should not fail anymore, but as this has not been extensively tested I prefer to raise the exception, because the users get confused  when no output is generated but the protocol finishes well and "in green". Check voxel size in _validate is a good idea and I have implemented it, however different originsmay be allowed.

DStrelak (approved):

DStrelak (approved):

PR #461, opened 2021-11-26
Shift Volume protocol: new protocol to shift volumes

EstrellaFG:
New protocol to shift volumes to a desired coordinate. It can be used to recenter a volume that has been computed from shifted particles.
Tests for shift volume and particles protocols are included.


EstrellaFG:
Once again, the remaining code smell is outside my domain 
Screenshot from 2021-11-29 10-07-24

albertmena:
> Once again, the remaining code smell is outside my domain Screenshot from 2021-11-29 10-07-24

Don't worry about this code smells, if you need that comment, leave it

PR #464, opened 2021-12-02
Continuousflex plugin: new basic tests

MohamadHarastani:
This will test the programs xmipp_nma_alignment, xmipp_nma_aligmment_vol and xmipp_volumeset alignment with their MPI versions. However, this test requires continuousflex plugin to be installed.

### DIFF
--- a/xmipp3/tests/test_protocols_continuousflex.py
+++ b/xmipp3/tests/test_protocols_continuousflex.py
@@ -1,0 +1,103 @@
+# **************************************************************************
+# * Authors:     Mohamad Harastani (mohamad.harastani@upmc.fr)
+# * IMPMC, Sorbonne University
+# *
+# * This program is free software; you can redistribute it and/or modify
+# * it under the terms of the GNU General Public License as published by
+# * the Free Software Foundation; either version 2 of the License, or
+# * (at your option) any later version.
+# *
+# * This program is distributed in the hope that it will be useful,
+# * but WITHOUT ANY WARRANTY; without even the implied warranty of
+# * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# * GNU General Public License for more details.
+# *
+# * You should have received a copy of the GNU General Public License
+# * along with this program; if not, write to the Free Software
+# * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA
+# * 02111-1307  USA
+# *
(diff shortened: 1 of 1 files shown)


DStrelak:
> However, this test requires continuousflex plugin to be installed.

@fonsecareyna82 Is this a problem / can we ensure that we have this plugin on Buildbot when running tests?

xmipp3/tests/test_protocols_continuousflex.py
```diff
@@ -0,0 +1,103 @@
...
+        self.launchProtocol(protSynthesize)
+        # ------------------------------------------------
+        # Lunch HEMNMA on the synthesized images (this will test xmipp_nma_alignment)
+        # ------------------------------------------------
+        protAlignment = self.newProtocol(FlexProtAlignmentNMA,
+                                         modeList='7-8')
+        protAlignment.inputModes.set(protNMA.outputModes)
+        protAlignment.inputParticles.set(protSynthesize.outputImages)
+        protAlignment.setObjLabel('HEMNMA backend')
+        self.launchProtocol(protAlignment)
```
DStrelak, line 103: So these test check only if the protocol finishes without errors, but do not verify the results. Is that correct?
If yes, can you verify results too?

DStrelak (changes requested):

xmipp3/tests/test_protocols_continuousflex.py
MohamadHarastani: Correct. I don't expect meaningful results from this test other than running the functions correctly. But I can make sure (with assert commands) that the output was correctly stored on the disk. I will take a look at what I can add next week.

DStrelak: Hi, any update?
If not, I would consider this PR ready, so that we have at least some tests for the release.

MohamadHarastani:
Hi David,
I have been very ill for a while, and I will be going on vacation for a
couple of weeks. Today is my last day of work this year and I have some
urgent stuff to do. You can consider it ready for the moment, I can improve
them later next year.

Happy new year in advance, and enjoy your holidays.
Mohamad

DStrelak:
Hi @MohamadHarastani 
I'm sorry to hear that. I hope you will get better soon!

I will proceed with merge.

Enjoy the holidays and see you soon!

David

DStrelak (approved):

PR #480, opened 2022-01-20
protocol_validate_fscq: output adapted to standard attributes viewer

DaniDelHoyo:
FSC-Q protocol output is AtomStruct.
Input AtomStruct as object

### DIFF
--- a/xmipp3/protocols/protocol_validate_fscq.py
+++ b/xmipp3/protocols/protocol_validate_fscq.py
@@ -25,7 +25,7 @@
 # *
 # **************************************************************************
 
-
+import shutil
 from pyworkflow import VERSION_3_0
 from pyworkflow.protocol.params import (PointerParam, BooleanParam,
                                         IntParam, FileParam, FloatParam)
@@ -84,7 +84,7 @@ class XmippProtValFit(ProtAnalysis3D):
 
         group.addParam('fromFile', BooleanParam, default=False,
                       label='Input PDB from file: ')
-        group.addParam('inputPDBObj', PointerParam, pointerClass='PdbFile', allowsNull=True,
+        group.addParam('inputPDBObj', PointerParam, pointerClass='AtomStruct', allowsNull=True,
                       label="Refined PDB: ", important=True, condition='not fromFile',
                       help='Specify the desired input structure.')
         group.addParam('inputPDB', FileParam,  condition='fromFile',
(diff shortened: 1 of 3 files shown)


DaniDelHoyo:
Los valores por defecto para el rango de fscq he puesto (-3,3), no se si habria que cambiarlo

erneyramirez (approved):

DStrelak (changes requested):
PR name does not match the content (this is clearly FSCQ related PR)

DaniDelHoyo:
> PR name does not match the content (this is clearly FSCQ related PR) 
Hi David, the PR adapts the viewer of the FSC-Q protocol to the general attributes viewer I developed, that's why the name.
If you think the name still must be changed, we'll do it

> Also please add tests for the FCSQ protocol. At least I don't see any at the buildbot.
do you have any test for the protocol @erneyramirez ?

erneyramirez:
@DStrelak I have to do the test, I had it pending. Pass the PR so Dani can get rid of it and I'll take care of the test

DStrelak:
> he PR adapts the viewer of the FSC-Q protocol to the general attributes viewer

In that case yes, please change the name of the PR (convention is roughly protocol_name: work_done)

DStrelak:
> @DStrelak I have to do the test, I had it pending. Pass the PR so Dani can get rid of it and I'll take care of the test

Please accompany this PR with the test directly. It's to avoid a situation in which we have not working code already in the devel.
Alternatively we can put this PR on hold, and you can create the test in a derived branch, eventually merging it to this PR and then to devel.

DaniDelHoyo:
There you have it, the test and the name is changed

xmipp3/protocols/protocol_validate_fscq.py
```diff
@@ -319,58 +335,58 @@ def runBlocresStep(self, i):
                         env=bsoft.Plugin.getEnviron(bsoftVersion=bsoft.V1_9_0))
```
DStrelak, line 335: the version of the program should be defined in a single place

DStrelak, line 335: I also get `Protocol failed: module 'bsoft' has no attribute 'V1_9_0'`

DStrelak (changes requested):
Test fails with
```
Error: Option -criterio is not defined!
Option errors: 1
Error: Option -criterio is not defined!
Option errors: 1
Traceback (most recent call last):
```

DStrelak (approved):

PR #498, opened 2022-04-05
Center particles: protocol simplified

albertmena:
Before the protocol to center particles had as output the centered classes, the centered particle and the coordinates.

With this PR the protocol is simplified and only centers the particle returning the classes and the particles, if the set of coordinates is needed,  you have to manually add the protocol to extract coordinates.


pconesa (approved):
Cleaner! Good job!

A minor optional, suggestion. It would be nice to start defining the possible outputs of each of the protocols. See root protocol._possibleOutputs comments. It's easy, I promise.

cossorzano (approved):

PR #502, opened 2022-04-11
Trigger data protocol: new function to stop the stream

danielmarchan3:
New functionality added to the trigger data protocol. Allows the connection of two trigger data protocols, the upstream protocol will be awaiting a signal from the downstream protocol. Upon reading this signal, the upstream protocol will stop the stream.


xmipp3/protocols/protocol_trigger_data.py
```diff
@@ -154,12 +170,22 @@ def _checkNewOutput(self):
...
+
         outputStep = self._getFirstJoinStep()
         deps = []
         if self.finished:  # Unlock createOutputStep if finished all jobs
             self._fillingOutput()  # To do the last filling
             if outputStep and outputStep.isWaiting():
                 outputStep.setStatus(cons.STATUS_NEW)
+                # Send the signal to the connected protocol
+                if self.triggerSignal.get():
+                    print('STOPPING MY OLDER BROTHER')
```
DStrelak, line 187: ?

albertmena (commented):
Review the help string. Replace * with /". Review also the text, "Send all particles to output" is not an option.

albertmena (approved):

PR #505, opened 2022-04-13
protocol_shift_particles: apply transform is now optional

EstrellaFG:
The computed shift was always applied to output particles, however, by users demand, now it is optional. Thus the program now can compute the shift and apply it or just store the computed shift in the metadata of the particle (transform matrix, x and y coordinates) but it does not apply it, which saves time. The stored shift could be applied afterwards using protocol "apply_alignment_2d" or re-extracting the particles with the shifted coordinates.

### DIFF
--- a/xmipp3/protocols/protocol_shift_particles.py
+++ b/xmipp3/protocols/protocol_shift_particles.py
@@ -61,6 +61,11 @@ class XmippProtShiftParticles(EMProtocol):
         form.addParam('inputMask', PointerParam, pointerClass='VolumeMask', label="Volume mask", allowsNull=True,
                       condition='not option', help='3D mask to compute the center of mass, the particles will be '
                                                    'shifted to the computed center of mass')
+        form.addParam('applyShift', BooleanParam, label='Apply shift to particles?', default='True',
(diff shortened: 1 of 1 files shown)


xmipp3/protocols/protocol_shift_particles.py
```diff
@@ -61,6 +61,11 @@ def _defineParams(self, form):
         form.addParam('inputMask', PointerParam, pointerClass='VolumeMask', label="Volume mask", allowsNull=True,
                       condition='not option', help='3D mask to compute the center of mass, the particles will be '
                                                    'shifted to the computed center of mass')
+        form.addParam('applyShift', BooleanParam, label='Apply shift to particles?', default='True',
+                      help='Yes: the output particles have shift applied. No: output particles have the computed shift '
```
DStrelak, line 65: How about: `Yes: The shift is applied to particle images and zero shift is stored in the metadata. No: The shift is stored in the transformation matrix in the metadata, but not applied to the particle image....`?

EstrellaFG: much nicer, thanks!

albertmena (approved):

PR #507, opened 2022-04-19
DenoiseParticles: fixing inputClasses bug 

albertmena:
With this PR the test _scipion3 tests xmipp3.tests.test_protocols_xmipp_2d.TestXmippDenoiseParticles_ is fixed.
Review that the change makes sense


cossorzano (approved):
I think it is correct

PR #527, opened 2022-06-03
Test: reconstruct fourier

DavidHerreros:
A new test was added to check that the execution of protocol `protocol_reconstruct_fourier.py` proceed as expected.

### DIFF
--- a/xmipp3/tests/test_protocol_reconstruct_fourier.py
+++ b/xmipp3/tests/test_protocol_reconstruct_fourier.py
@@ -1,0 +1,90 @@
+# ***************************************************************************
+# * Authors:     David Herreros Calero (dherreros@cnb.csic.es)
+# *
+# *
+# * This program is free software; you can redistribute it and/or modify
+# * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


Vilax (approved):

xmipp3/tests/test_protocol_reconstruct_fourier.py
```diff
@@ -0,0 +1,82 @@
...
+        """ Check that an ouput was generated and
+        the condition is valid.
+        """
+        o = getattr(prot, outName, None)
+        locals()[outName] = o
+        self.assertIsNotNone(o, "Output: %s is None" % outName)
+        for c in conditions:
+            self.assertTrue(eval(c), 'Condition failed: ' + c)
+
+    def createPhantomSphere(self):
```
DStrelak, line 48: from my experience, sphere is not the best geometrical object to test this

DavidHerreros: I have modified the geometry of the target volume from a sphere to a cube, not sure if this way reconstruction should be better

cossorzano (approved):

PR #535, opened 2022-06-13
Protocol Analyze Local CTF: bugfixing + test + formatting

EstrellaFG:


xmipp3/protocols/protocol_analyze_local_ctf.py
```diff
@@ -143,8 +145,7 @@ def createOutputStep(self):
         self._defineSourceRelation(self.inputSet, outputSet)
         self._defineSourceRelation(inputMicSet, outputSet)
 
-
-    #--------------------------- INFO functions --------------------------------------------
+    # --------------------------- INFO functions --------------------------------------------
     def _summary(self):
```
albertmena, line 149: You could add a median of the defocus (how to: search summary.txt files), to have a general idea of how it was.

xmipp3/viewers/viewer_analyze_local_ctf.py
```diff
@@ -56,37 +54,36 @@ def getData(self):
     def _defineParams(self, form):
         form.addSection(label='Visualization')
         form.addParam('displayR2', LabelParam, default=False, label='Display micrograph R2')
-        form.addParam('displayLocalDefocus', IntParam, default=1,
-                      label='Display local defocus of micrograph:',
+        form.addParam('displayLocalDefocus', IntParam, label='Display local defocus of micrograph:', allowsNull=True,
```
albertmena, line 57: Help identifying the ID of the available micrographs. Or at least ad by default the first (default=1?)

albertmena (changes requested):

xmipp3/protocols/protocol_analyze_local_ctf.py
EstrellaFG: In this case we have a defocus value for each of the particles and thus has no sense to have a "global" defocus value

xmipp3/tests/test_protocols_local_defocus.py
```diff
@@ -105,3 +104,41 @@ def testXmippLocalDefocusConsensus(self):
...
+        protAnalyzeLocalDefocus = self.newProtocol(XmippProtAnalyzeLocalCTF,
+                                                   inputMics=protImportMics.outputMicrographs,
+                                                   inputSet=protImportParticles.outputParticles)
+        self.launchProtocol(protAnalyzeLocalDefocus)
+        self.assertIsNotNone(protAnalyzeLocalDefocus.outputMicrographs.getFiles(),
+                             "There was a problem with analysis of local CTF estimation")
+        self.assertEqual(protAnalyzeLocalDefocus.outputMicrographs.getDim(), (9216, 9441, 1),
+                         "There was a problem with size of set of particles after consensus")
+        self.assertEqual(protAnalyzeLocalDefocus.outputMicrographs.getFirstItem().getSamplingRate(), 1.237,
+                         "There was a problem with the sampling rate value of output consensus particles")
```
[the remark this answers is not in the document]
EstrellaFG: What it is computed in this protocol is an R2 coefficient that "compares" the defocus value of the micrograph with the defocus of each particle after the local refinement, but that R2 coeffcient it is not an attribute of the micrograph class, but a new xmipp label added to the xmipp metadata, so it is not accesible through the test (AFAIK)

xmipp3/viewers/viewer_analyze_local_ctf.py
```diff
@@ -56,37 +54,36 @@ def getData(self):
...
-                mdPoints = emlib.MetaData("mic_%d@%s"%(self.displayLocalDefocus.get(),fnDefoci))
+                mdPoints = emlib.MetaData("mic_%d@%s" % (self.displayLocalDefocus.get(), fnDefoci))
 
                 x = mdPoints.getColumnValues(emlib.MDL_XCOOR)
                 y = mdPoints.getColumnValues(emlib.MDL_YCOOR)
                 defocusA = mdPoints.getColumnValues(emlib.MDL_CTF_DEFOCUSA)
                 residuals = mdPoints.getColumnValues(emlib.MDL_CTF_DEFOCUS_RESIDUAL)
 
-                title="Micrograph %d defocus"%self.displayLocalDefocus.get()
+                title="Micrograph %d defocus" % self.displayLocalDefocus.get()
```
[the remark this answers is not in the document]
EstrellaFG: We need a 3D graph because the defocus value for each particle "is equivalent to the height of the particle in the ice", and that is exactly what we want to see in the plot. We want to see both the "real" height for each particle (Avg = average between defocusU and defocusV) and the adjusted plane to see the plane where all particles "should be" if they have equal defocus. The range of the defocus axis can be modified dynamically in the plot, which is usefull as you may want to compare different runnings of this protocol with different defocus estimations as input, and they will produce outputs with different ranges, so you can adjust them dynamically.

xmipp3/protocols/protocol_analyze_local_ctf.py
albertmena: ok, and maybe the R2?, I dont know if is an interesting parameter to view in the summary

albertmena (approved):

PR #549, opened 2022-07-14
protocol_deep_hand: new test for the protocol

gemapatricia:


EstrellaFG (dismissed):

xmipp3/protocols/protocol_deep_hand.py
```diff
@@ -41,6 +41,9 @@ class XmippProtDeepHand(EMProtocol, XmippProtocol):
 
     _label ="deep hand"
     _conda_env = "xmipp_deepHand"
+    vResizedVolFile = 'resizedVol.mrc'
```
albertmena, line 44: define this params inside the class, _insertAllStep needs functions without params as input.

albertmena (changes requested):
Could you add a validation in this protocol?, just add 

>errors.append( self.validateDLtoolkit()))
in the validate function
thanks

xmipp3/protocols/protocol_deep_hand.py
albertmena: Also merge devel into current branch to fix the conflict

xmipp3/protocols/protocol_deep_hand.py
```diff
@@ -169,5 +169,6 @@ def _validate(self):
            errors.append("Alpha threshold must be between 0.0 and 1.0")
         if self.thresholdHand.get() > 1.0 or self.thresholdHand.get() < 0.0:
            errors.append("Hand threshold must be between 0.0 and 1.0")
+        errors.append(self.validateDLtoolkit())
```
EstrellaFG, line 172: @albertmena is it possible to add a more readable error message ? In my case (i do not have the deeplearning toolkit) I am getting: `'Protocol has validation errors:\n' + '\n'.join(errors))
00006:   TypeError: sequence item 0: expected str instance, list found`, which I guess is correct but I think for the user is better something like "deepLearningToolkit missing" or something similar

xmipp3/tests/test_protocols_xmipp_3d.py
```diff
@@ -1760,6 +1765,50 @@ def testXmippAlignVolumeAndParticles(self):
                          protCreatePhantomReference.outputVolume.getSamplingRate(),
                          (MSG_WRONG_SAMPLING, "rotated phantom"))
 
+class TestXmippDeepHand(TestXmippBase):
+    """This class checks if the protocol deep hand in Xmipp works properly."""
+    @classmethod
+    def setUpClass(cls):
+        setupTestProject(cls)
+        cls.dataset = DataSet.getDataSet(db_xmipp_tutorial)
+        cls.vol1 = cls.dataset.getFile(vol1_iter2)
```
EstrellaFG, line 1774: remove this line and the line before as they are no longer needed

xmipp3/protocols/protocol_deep_hand.py
```diff
@@ -165,5 +167,6 @@ def _validate(self):
            errors.append("Alpha threshold must be between 0.0 and 1.0")
         if self.thresholdHand.get() > 1.0 or self.thresholdHand.get() < 0.0:
            errors.append("Hand threshold must be between 0.0 and 1.0")
+        #errors.append(self.validateDLtoolkit(model="deepHand"))
```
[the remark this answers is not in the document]
gemapatricia: Leaving it commented for the moment, because if it is uncommented, the protocol doesn't work

xmipp3/tests/test_protocols_xmipp_3d.py
gemapatricia: Done

EstrellaFG (dismissed):

EstrellaFG (approved):

albertmena (approved):

PR #555, opened 2022-07-29
Vi dir align

Vilax:


Vilax:
The output of resolutoin_bfactor will be treated in a different PR. It can be done in a fancy way

gemapatricia (approved):

PR #571, opened 2022-09-30
Test for protocol resolution b factor 

EstrellaFG:
I have checked with Vilas and this test for protocol resolution b factor is ready and working

### DIFF
--- a/xmipp3/tests/test_protocols_xmipp_3d.py
+++ b/xmipp3/tests/test_protocols_xmipp_3d.py
@@ -1873,9 +1873,7 @@ class TestXmippResolutionBfactor(TestXmippBase):
                                                  localResolutionMap=protCreateMap.resolution_Volume,
                                                  fscResolution=8.35)
         self.launchProtocol(protbfactorResolution)
-        # Check if there is an output atomic structure
-        #self.assertIsNotNone(protbfactorResolution.getFiles(),
-        #                     (MSG_WRONG_PROTOCOL, "local resolution/local bfactor"))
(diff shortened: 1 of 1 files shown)


xmipp3/tests/test_protocols_xmipp_3d.py
```diff
@@ -1807,6 +1813,70 @@ def testXmippDeepHand(self):
...
+
+        # Protocol local resolution/local bfactor
+        print("Protocol local resolution/local bfactor")
+        protbfactorResolution = self.newProtocol(XmippProtbfactorResolution,
+                                                 pdbfile=protImportPdb.outputPdb,
+                                                 localResolutionMap=protCreateMap.resolution_Volume,
+                                                 fscResolution=8.35)
+        self.launchProtocol(protbfactorResolution)
+        # Check if there is an output atomic structure
+        #self.assertIsNotNone(protbfactorResolution.getFiles(),
```
albertmena, line 1877: why is commented the assertIsNotNone?

albertmena (changes requested):

xmipp3/tests/test_protocols_xmipp_3d.py
EstrellaFG: Because we where not sure if an output was produced by the protocol or not, but last Friday I checked with Vilas that no output is produced (the "output" of this protocol is just the viewer). Thus, I have already removed the commented code :)

albertmena (approved):

PR #580, opened 2022-10-21
Update README.md

albertmena:


danielmarchan3:
When installing with scipion installation we should tell them to change the flag -noXmipp in step 4 of the guide.

danielmarchan3 (approved):

PR #581, opened 2022-10-21
Fix bug: Trigger data protocol

danielmarchan3:
Add a new validation to the trigger data protocol

### DIFF
--- a/xmipp3/protocols/protocol_trigger_data.py
+++ b/xmipp3/protocols/protocol_trigger_data.py
@@ -174,14 +174,14 @@ class XmippProtTriggerData(EMProtocol):
 
         # Send the signal to the connected protocol
         if self.triggerSignal.get() and len(self.images) >= self.outputSize:
-            print('Sending signal to stop the input trigger data protocol')
+            self.info('Sending signal to stop the input trigger data protocol')
             self.stopWait()
 
         # Wait for trigger data signal
         if self.triggerWait.get():
-            print('Waiting for signal to stop the stream')
+            self.info('Waiting for signal to stop the stream')
             if self.waitingHasFinished():
-                print('Stopped by received signal from a trigger data protocol')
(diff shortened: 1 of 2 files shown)


xmipp3/protocols/protocol_trigger_data.py
```diff
@@ -94,7 +94,7 @@ def _defineParams(self, form):
                            '\n For this option, select the option send all items to output.')
         form.addParam('triggerProt', PointerParam,
                       pointerClass=self.getClassName(),
-                      condition='triggerSignal', allowsNull=True,
+                      condition='triggerSignal',
```
albertmena, line 97: Trigger data protocol does not appear in the GUI, is that correct?

xmipp3/protocols/protocol_trigger_data.py
```diff
@@ -173,10 +173,9 @@ def _checkNewOutput(self):
             self.finished = False
 
         # Send the signal to the connected protocol
-        if self.triggerSignal.get():
-            if len(self.images) >= self.outputSize:
-                print('Sending signal to stop the input trigger data protocol')
-                self.stopWait()
+        if self.triggerSignal.get() and len(self.images) >= self.outputSize:
+            print('Sending signal to stop the input trigger data protocol')
```
albertmena, line 177: better self.info() than print()

xmipp3/protocols/protocol_trigger_data.py
albertmena: Sorry, What I want to say is: 
" triggerProt" parameter does not appear in the 
Screenshot from 2022-10-28 11-50-54
window

danielmarchan3: That is correct, it only appears when the Send signal to a stop a stream is true.

albertmena (approved):

PR #596, opened 2022-12-01
Replacing outPbd to outPdb

albertmena:


jamesmkrieger (changes requested):
It’s needed further down too: line 128 too

jamesmkrieger:
We should probably add a test for this case too, maybe on both xmipp and plugin sides

DStrelak:
> We should probably add a test for this case too, maybe on both xmipp and plugin sides

Yeah, problem is that this protocol does not have any test whatsoever and I'm currently overwhelmed with other work :/

albertmena:
it has a test here: https://github.com/I2PC/scipion-em-xmipp/blob/6098a3a7d1df2b170849ec41e67cd71267dca387/xmipp3/tests/test_protocols_xmipp_3d.py#L1118

PR #605, opened 2023-01-11
Fixed a documentation issue

MohamadHarastani:


MohamadHarastani:
Hi @EstrellaFG 
Would you please verify that this was a problem that I solved and not introduced :-)

Cheers
Mohamad

EstrellaFG:
Hi, I think this should be checked better by @cossorzano

MohamadHarastani:
Alright! Hi Carlos.
I believe there is a bug in the Xmipp code (xmipp_transform_filter), it says A for defocus but actually requires nm.
I have never used this way to simulate CTF, but I checked the other way that I have been using for a while to simulate CTF using xmipp_phantom_simulate_microscope. There, we need to provide "ctf.param" file and give the defocus in nm too (although it is not documented), but unlike here, negative is defocus and positive is overfocus.
We have a protocol to synthesize images based on xmipp_phantom_simulate_microscope, here is the function that calls it
https://github.com/scipion-em/scipion-em-continuousflex/blob/master/continuousflex/protocols/protocol_image_synthesize.py#L560

EstrellaFG (changes requested):
Hi Mohammad, sorry for the delay. I have checked it with COSS and the correct unit is angstroms, so please, keep the "A". But the comment about overfocus is fine and usefull, thanks.

MohamadHarastani:
Thanks @EstrellaFG , I will take another look and be back to you.

MohamadHarastani:
Hello @EstrellaFG 
You are right. I reverted the changes and kept only the comment.
Cheers
Mohamad

PR #611, opened 2023-02-06
protocol_projection_subtraction: user experience improvements

EstrellaFG:
- Add warning to change parameter pad to 1 if particles are quite big, in order to save RAM and avoid core dumped
- Create a viewer in order to automatically display R2subtraction column and hide unnecessary columns when clicking in analyze results


EstrellaFG:
The code smell reported by sonar cloud is not related with this PR

albertmena (approved):

PR #616, opened 2023-02-27
- Align volume and particles works in the absence of tomo plugin.

pconesa:
Without tomo plugins the align volume and particles was failing. This will tolerate the absence of tomo.

NOTE: Hot fix to master.... needs PR to devel.


albertmena:
In one week I plan to create a new release, maybe is not necessary a hotfix?

albertmena (approved):

PR #630, opened 2023-03-21
protocol_subtract_projection: change pad validation error for warning

EstrellaFG:
hot fix: change pad validation error for warning  to be able to execute the protocol with big particles


albertmena:
If you consider that this changes must be in the master branch (and available for users that work with the production version (no devel) please create a PR from your branch to master

albertmena (approved):

PR #645, opened 2023-03-31
CTF consensus: add 4 threads by default

danielmarchan3:
We tested the parallelization of this protocol using threads and mpis and we did not see any issue.


xmipp3/protocols/protocol_ctf_consensus.py
```diff
@@ -658,10 +658,8 @@ def compareValue(ctf, label, comp, crit):
 
             if self.xmippCTF == INPUT1:
                 ctfX = ctf
-            elif self.xmippCTF == INPUT2:
```
albertmena, line 661: Why you remove INPUT2 option?

albertmena (approved):

PR #648, opened 2023-04-05
test continuousFlex fixed

albertmena:
Now the dataset is created/downloaded/checked at the begin of the test


Vilax (approved):

MohamadHarastani:
hello @Vilax and @albertmena 
Sorry I have just gotten back from vacations. I think the fix is fine. I am only concerned that updating the dataset later will require redefining it in two places. I will keep that in mind later and think of a fix.
Thank you both!
Mohamad

PR #657, opened 2023-04-25
Update requirements.txt

albertmena:


pconesa (changes requested):
Hi Alberto!

The current released package in pypi has the same issue.

I think it is better to fix this in production (branching out from master and bumping the version) and then sending the hotfix to devel.

albertmena:
> Hi Alberto!
> 
> The current released package in pypi has the same issue.
> 
> I think it is better to fix this in production (branching out from master and bumping the version) and then sending the hotfix to devel.

Ok: https://github.com/I2PC/scipion-em-xmipp/pull/659

PR #668, opened 2023-05-04
protocol_consensus_local_ctf: save defocus in proper fields

EstrellaFG:
consensus local defocus estimation was saved in an xmipp field and thus, following reconstruction programs did not read that value when computing reconstructions. Thus, now the consensus local defocus estimation is saved directly in fileds defocusU and defocusV of the generic ctfModel.


pconesa (commented):
Changes are find.

Id take this chance to:

1.- add possible outputs
2.- use methods instead of strings in insertFunctionStep.

pconesa:
3.- add a line in changes.txt

pconesa (approved):

PR #669, opened 2023-05-04
protocol_create3dMaks: add :mrc to input filename

EstrellaFG:


cossorzano (approved):

xmipp3/protocols/protocol_preprocess/protocol_create_mask3d.py
```diff
@@ -208,6 +208,8 @@ def _insertAllSteps(self):
     def createMaskFromVolumeStep(self):
         volume = self.inputVolume.get()
         fnVol = getImageLocation(volume)
+        if fnVol.endswith(".mrc"):
+            fnVol += ":mrc"
```
pconesa, line 212: Guys! we should avoid adding :mrc to the paths.

This is usually needed because the heders of the mrc are wrong. In this case Volume class has a method fixMRCVolume to do so and avoid the problematic and tedious handling of ":mrc" in later protocols.

PR #673, opened 2023-05-08
AlignVolume and viewer structMap

albertmena:
- Included the label in the volumes
- Change the label for each volume


albertmena:
Not sure about the labelling of crop/resize protocol, looks like create a label for each vol and in my case overwrite the manual labelling

DavidHerreros (approved):

PR #674, opened 2023-05-08
updating devel with master

pconesa:


EstrellaFG:
We have mergued PR https://github.com/I2PC/scipion-em-xmipp/pull/660 with these changes in requirements and version. On the other hand, the protocol_subtract_projection in devel is updated and it will be updated in master when release is done

PR #693, opened 2023-06-07
Add mask as input. Mask resize is now possible.

erneyramirez:


xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -265,16 +276,42 @@ def resizeStep(self, isFirstStep, args):
     
     def windowStep(self, isFirstStep, args):
         XmippResizeHelper.windowStep(self, self._ioArgs(isFirstStep)+args)
+    
+    def convertInputStep(self):
+        """ convert if necessary"""
+        if self.inputParticles.get().getClassName() == 'Mask':
```
pconesa, line 282: I'd extract this into a singe method 

def isMask():

   return isInstance(self.inputParticles.get(), Mask)

pconesa: and of course use it everyowhere

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -265,16 +276,42 @@ def resizeStep(self, isFirstStep, args):
...
-        self.inputHasAlign = self.inputParticles.get().hasAlignment()
+        isMask = True if self.inputParticles.get().getClassName() == 'Mask' else False
+        if not isMask:
+            self.inputHasAlign = self.inputParticles.get().hasAlignment()
         
         if self.doResize:
             output.setSamplingRate(self.samplingRate)
-            setMRCSamplingRate(self.outputStk, self.samplingRate)
+            if not isMask:
+                setMRCSamplingRate(self.outputStk, self.samplingRate)
```
pconesa, line 314: I'd also set de samplingrate of the mask n the mrc file too, providing is an mrc.

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
MartinSalinas98: Done

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
MartinSalinas98: Hi @pconesa, what do you mean with this? @erneyramirez and I have tried this with a mask which has a .xmd (or similar) extension, which does not work with that function, and we are also doing always `output.setSamplingRate(self.samplingRate)`.
Does that make any difference, and if so, what should we do? Only not run that function for non-mrc Masks, or always try to convert a non-mrc Mask into an mrc?

pconesa: Please, if possible, better generate mrc files. Those functions only apply to mrc files.

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -244,11 +246,22 @@ class XmippProtCropResizeParticles(XmippProcessParticles):
     """ Crop or resize a set of particles """
     _label = 'crop/resize particles'
     _inputLabel = 'particles'
+    _possibleOutputs = {'outputParticles': SetOfParticles, 'outputMask': Mask}
```
pconesa, line 249: Nice to define possible outputs.

One  suggestion, further down you use "outputMask" again in the defineOutputs. We usually define a constant and use the constant everywhere. Even in tests.

To avoid hasAttr....you can define in the __init__ a None Attribute like self.outputParticles (same for outputMask)

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -265,12 +278,40 @@ def resizeStep(self, isFirstStep, args):
...
+    def convertInputStep(self):
+        """ convert if necessary"""
+        if self.isMask(self.inputParticles.get()):
+            # If input is a Mask, modify filter params
+            self.inputFn = self.inputParticles.get().getFileName()
+            inputName = os.path.splitext(os.path.basename(self.inputFn))[0]
+
+            # Set output mask path
+            self.outputStk = self._getExtraPath(os.path.basename(inputName + '.mrc'))
+            self.outputMd = self._getTmpPath('tmp.xmd')
```
pconesa, line 291: this extension is not right. And there is no conversion here in the case of the mask..it is just filling attributes. Looks confusing.

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -341,6 +382,14 @@ def _validate(self):
         return XmippResizeHelper._validate(self)
     
     #--------------------------- UTILS functions ---------------------------------------------------
+    def isMask(self, inputObject):
```
pconesa, line 385: Why passing a argument it can get from self?

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -341,6 +382,14 @@ def _validate(self):
...
+      """ This function returns True if the given object is a Mask. False otherwise. """
+      try:
+        return inputObject.getClassName() == 'Mask'
```
pconesa, line 388: Maybe isInstance will work and avoid AttributeError and the try/except

pconesa (changes requested):
I advice some minor changes.

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
MartinSalinas98: This is because outputMd is used for the metadata file when working with a set of particles. In this case, we redirect it to tmp path so it gets deleted, because we dont need it. We could also override the parent functions where that attribute is referenced and avoid using it when working with masks, but this was way simpler

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
MartinSalinas98: True, done

xmipp3/protocols/protocol_preprocess/protocol_crop_resize.py
MartinSalinas98: Done

pconesa (approved):

PR #694, opened 2023-06-12
Update binary DLTK version

albertmena:
Update binary DLTK version
Add installation of DLTK in the action


fede-pe (dismissed):

oierlauzi (approved):

MartinSalinas98:
DLTK cannot be installed in a Github hosted action, as Nvidia Drivers need to be installed as a dependency of DLTK, and that is not possible in the Github worker, neither rebooting afterwards.

To test the installation of DLTK, integration with a self-hosted machine will be necessary.

MartinSalinas98 (approved):

PR #702, opened 2023-06-30
added a protocol to convert a set of pdbs to volumes

MohamadHarastani:
Hello
I created a simplified version of "convert a pdb" that takes a set of atomic structures and converts it to a set of volumes. While trying to do that, I noticed that the options for importing a PDB from a file or by ID directly to "convert a pdb" are not too useful, so I skipped them to avoid complicating the code. These options are allowed by "import a pdb", so I don't see a need to maintain them at two different places.
Some other options that can be useful when converting a single PDB to volume are not too important when converting a set of PSBs, so I skipped them. But let me know if you have other thoughts.
Cheers
Mohamad


MohamadHarastani:
btw, It has been a while since I compiled Xmipp, and I noticed that it has been behaving better than before. Good job!

MartinSalinas98:
Hey @MohamadHarastani, thank you for your contribution!
I mostly agree with your ideas, but I have a couple of suggestions.

Would it be possible, instead of creating a new protocol, modify the necessary code in the "convert a pdb" protocol? That is because, we are always trying to avoid having multiple protocols doing the same or very similar things (even if it ends up happening sometimes anyway).

Because of that modification, then it would be necessary to keep the parameters specific for single atomic structure conversion (but they can be shown/hidden in the form deppending on the type of the input, hiding them when working with sets if needed).

And I think we can safely remove the import from ID or file since it's true that such functionality is already present in the import protocol.

If you could change that it would be great, and let me know if you have any question about it or further suggestions!

MohamadHarastani:
Hello @MartinSalinas98 
Fair enough! I'll get back..

MartinSalinas98 (changes requested):
Waiting for modification: adapt convert a pdb instead of creating new protocol

MartinSalinas98:
I was trying to test this protocol locally, but I have realised that import atomic structure does not support importing several atomic strcutures. Were you able to test your changes?

MohamadHarastani:
Indeed, import atomic structure does not allow importing a set. There's another protocol to import a set of atomic structures.
I created a test protocol, you can run "scipion3 tests xmipp3.tests.test_protocols_xmipp_3d.TestXmippPdbConvert"

MartinSalinas98:
> Indeed, import atomic structure does not allow importing a set. There's another protocol to import a set of atomic structures. I created a test protocol, you can run "scipion3 tests xmipp3.tests.test_protocols_xmipp_3d.TestXmippPdbConvert"

Thanks for the info about the set import protocol, I hadn't noticed!
I have reviewed the code, and I have seen some duplication. With that I mean, the code for single atom structs and sets is very similar (which makes sense, because it should be almost the same but the on for sets in a loop). I am currently working on fusing both convert steps and also fusing the code of convert output to reduce code duplication, maybe tomorrow it will be ready, and I'll let you know as soon as it is.

MohamadHarastani:
Great job! Since you got yourself involved, could you parallelize the loop too?

MartinSalinas98:
> Great job! Since you got yourself involved, could you parallelize the loop too?

I don't think it makes a lot of sense in this situation. Let me explain:
- One possibility is to check if the xmipp program below this protocol (the one being called with `runJob`) has already been parallelized with mpi, which I don't think it is, because each run takes in the neighborhood of 2s, so it would have on one side, nothing to parallelize (because it performs one conversion per call), and, on the other side, absolutely no benefits, you are very likely not going to improve the 2s conversion with pararllelization. So this option is out of question because it is probably not possible, and, if it was, it would very likely degrade performance in this case, a whole restructuring of the xmipp C++ code for this program would be necessary to set it in a way that can be parallelized and gain performance with it.
- The second option is to use the scipion native `mpi4py` API, but this is not a good idea, not in this case but in general, because scipion is going to drop support for this in the near future (in summary, it usually caused installation issues frecuently, was rarely used, and some other issues that I'm not that familiar with, but more experienced scipion developers will probably know about).
- The last option is an implementation of some sort of parallelization of our own inside the protocol itself. This, although doable, is problematic due to the nature of the sets, because their elements need to be inserted in the same order of the input set of atom structs, or the user would have lost that reference (and, as you might probably know, throwing several processes with different inputs, does not guarantee the will always finish in the same order, in fact, they likely won't in most cases). Controlling the same order is fulfilled at the end is possible, but an extra task that also has to be done ad-hoc for this protocol, and, taking into account that each conversion takes about 2s, this would only be truly noticeable when the input set contains thousands of atom structs. I can try to take a look into this, but I can't promise I'll actually implement it since its a pretty niche use case and an extra development effort that won't be useful for most users.

jamesmkrieger:
> > Great job! Since you got yourself involved, could you parallelize the loop too?
> 
> I don't think it makes a lot of sense in this situation. Let me explain:
> 
> * One possibility is to check if the xmipp program below this protocol (the one being called with `runJob`) has already been parallelized with mpi, which I don't think it is, because each run takes in the neighborhood of 2s, so it would have on one side, nothing to parallelize (because it performs one conversion per call), and, on the other side, absolutely no benefits, you are very likely not going to improve the 2s conversion with pararllelization. So this option is out of question because it is probably not possible, and, if it was, it would very likely degrade performance in this case, a whole restructuring of the xmipp C++ code for this program would be necessary to set it in a way that can be parallelized and gain performance with it.
> * The second option is to use the scipion native `mpi4py` API, but this is not a good idea, not in this case but in general, because scipion is going to drop support for this in the near future (in summary, it usually caused installation issues frecuently, was rarely used, and some other issues that I'm not that familiar with, but more experienced scipion developers will probably know about).
> * The last option is an implementation of some sort of parallelization of our own inside the protocol itself. This, although doable, is problematic due to the nature of the sets, because their elements need to be inserted in the same order of the input set of atom structs, or the user would have lost that reference (and, as you might probably know, throwing several processes with different inputs, does not guarantee the will always finish in the same order, in fact, they likely won't in most cases). Controlling the same order is fulfilled at the end is possible, but an extra task that also has to be done ad-hoc for this protocol, and, taking into account that each conversion takes about 2s, this would only be truly noticeable when the input set contains thousands of atom structs. I can try to take a look into this, but I can't promise I'll actually implement it since its a pretty niche use case and an extra development effort that won't be useful for most users.

This last option could be done by inserting them in a separate loop after collecting all the results. We could return a tuple containing an id as well as a map and loop through those.

2s per structure is actually a long time when we have thousands of structures or more

MartinSalinas98:
@jamesmkrieger I know, it adds up with such big amounts of structures, but i've never seen anyone use sets of atomic structures with that many elements. If you think it's something to take into account, sure, i can try to implement now that I'm at it.

jamesmkrieger:
I think we probably will have those kinds of numbers of structures although probably we should use some different kind of set with a file that only contains coordinates and not all the atom information in that case because otherwise we need a lot of disk space and i/o and I am working on such an object. 

We can then have another protocol to temporarily convert the coordinate set entries to pdb files and delete them as we go or something. 

That’s something longer term though. It could be that continuousflex or zernike3d or another continuous flexibility method together with phenix varref may generate a large number of pdb files that people may want to convert although it’s recommended to do clustering first anyway. I guess we could still get into the order of 100 cluster representatives though

MohamadHarastani:
Hello @jamesmkrieger 
it's been a while! hope you're fine.
I agree on the last point. We can still parallelize the loop, by breaking the pdb set object and re-making it after the conversion to volumes is done. I have done that in my initial code, I was collecting the file names separately, and then building a set of volumes in the same order of the set of pdbs. I didn't look at the modified code by Martin.
There are many applications that will build up, but speed is not a major problem at the moment.
@MartinSalinas98 Feel free to skip this parallelization now, we will address it later if needed.
Cheers

MartinSalinas98:
As of right now, I have removed duplication within the code. I have talked to Pablo, and there might be a chance that parallelization is possible in a controlled way within scipion without having to manually implement it, and the structure of the code will vary deppending on if the parallelization takes place in the isert steps function or within the processing function. So I'll wait until I know more about it to implement any of it.

MartinSalinas98 (dismissed):

jamesmkrieger:
Ok, sounds good :)

MohamadHarastani:
> As of right now, I have removed duplication within the code. I have talked to Pablo, and there might be a chance that parallelization is possible in a controlled way within scipion without having to manually implement it, and the structure of the code will vary deppending on if the parallelization takes place in the isert steps function or within the processing function. So I'll wait until I know more about it to implement any of it.

Alright, makes sense. I found however a bug when converting a set of pdbs to volumes. The size is not taken properly from the user.

MohamadHarastani:
I committed a fix, but somehow I don't see it in the PR
https://github.com/I2PC/scipion-em-xmipp/commit/11cd30093b862a2635bb24c68002a73cfaa4e1a1

MartinSalinas98:
> I committed a fix, but somehow I don't see it in the PR [11cd300](https://github.com/I2PC/scipion-em-xmipp/commit/11cd30093b862a2635bb24c68002a73cfaa4e1a1)

Me neither here, but I can certainly pull them. I pushed them again with a slight modification (removing `pass` instruction, since it does nothing and is not needed).

MartinSalinas98 (approved):

MohamadHarastani:
> Added new commit removing an extra line of unused code (not related to your commit). Restarting checks. It looks like it takes a while to load, like your previous commit, I'm pretty sure is GitHub acting up, it should be eventually solved.

Alright! so we wait :-)

MartinSalinas98 (dismissed):

MohamadHarastani:
Another improvement. When we choose the option to store the centered pdbs, these pdbs are deleted anyway when the protocol is finished. They should be kept, and imported to scipion as an object. I'll fix that before we merge this PR.

MohamadHarastani:
Hi @MartinSalinas98 
This is great! I will need to work on this improvement this afternoon. I will try to catch up with the release. 
Cheers

albertmena (approved):
Great job!

albertmena (approved):

MartinSalinas98:
@MohamadHarastani thanks for the change! I have pushed a new commit that includes a small detail you left out. Now the centered pdbs have their `volume` property set to the volume they come from, so they are linked within scipion.

MartinSalinas98 (approved):

albertmena:
Thanks all for the great and fast work!

PR #705, opened 2023-07-04
DeepCenter: New protocols

albertmena:
4 new protocols that performs centering and global assignment of  a set of particles:
- deep_center: uses a labeled set of particles to train CNN-model(s) to center particles
- deep_center_predict: uses input trained model(s) to predict centers of an input set of particles
- deep_global_assignment: uses a labeled set of particles to train CNN-model(s) to perform global assignment of Euler angles
- deep_global_assignment_predict: uses input trained model(s) to assign angles of an input set of particles


albertmena:
Two comments:
A test is required for each protocol
Please try to run the programs with [xmipp_DLTK_v1.0.yml](https://github.com/I2PC/xmipp/blob/devel/src/xmipp/bindings/python/envs_DLTK/xmipp_DLTK_v1.0.yml) instead of [xmipp_DLTK_v0.3.yml](https://github.com/I2PC/xmipp/blob/devel/src/xmipp/bindings/python/envs_DLTK/xmipp_DLTK_v0.3.yml).

AdrianSansi (approved):

PR #717, opened 2023-09-13
The model can now be given by a directory

cossorzano:
The picking model can now be given from a directory instead of a previous manual picking. This is very useful in cases where the same macromolecules are imaged in many different projects.


danielmarchan3 (approved):
I tested in two different projects and it works fine.

PR #722, opened 2023-09-29
Test Mixed Movies: fixed

danielmarchan3:
Remove some tests cases that were no making sense. 

Movie alignment from a determined frame range was correctly done but not correctly assessed.


albertmena (changes requested):
The test still failling:

image

albertmena:
Maybe changing the tolerance?

albertmena (approved):
Just fails the test in buildbot not all machines

PR #723, opened 2023-09-29
Tilt analysis protocol: fixes in the generated tilt images and in updating correctly the output sets

danielmarchan3:
- Remove the extra empty space in the generated tilt images
- Update correctly the output sets using scipions generic method


albertmena:
It fixes any test or issue? please inform about it

danielmarchan3:
> It fixes any test or issue? please inform about it

The test does not catch this error because the output is updated in the database but not in the Scipion object. If you run this protocol, you would notice that in the output, only the first batch (N) appears. However, if you open the output, you will see that all the images are displayed, despite it stating that the output contains N movies.

albertmena (approved):

PR #724, opened 2023-09-29
Movie gain protocol: fix update output step by using a generic one from scipion

danielmarchan3:
- Use _updateOutputSet() generic method from class Protocol of scipion-pyworkflow. This method is used for streaming protocols.


MartinSalinas98 (approved):

albertmena:
It fixes any test or issue? please inform about it

danielmarchan3:
> It fixes any test or issue? please inform about it

The test does not catch this error because the output is updated in the database but not in the Scipion object. If you run this protocol, you would notice that in the output, only the first batch (N) appears. However, if you open the output, you will see that all the images are displayed, despite it stating that the output contains N movies.

PR #737, opened 2023-10-26
Deprecating GL2D

albertmena:


danielmarchan3 (changes requested):
Delete as well from the protocols.conf (XmippProtStrGpuCrrCL2D, etc...)

danielmarchan3 (approved):

PR #740, opened 2023-10-30
Particle picking consensus: fix MicsPointer

danielmarchan3:
The image pointer for the micrographs was assuming a particular name for the micrographs pointer and when they were dose weighted it does not fit.


pconesa (approved):
Easier!!

PR #756, opened 2023-12-13
Added support for reading EER movies to FlexAlign

oierlauzi:
Requires https://github.com/I2PC/xmippCore/pull/197 to be merged first
Adds EER processing capabilities to FlexAlign. Other movie processing protocols remain pending


albertmena:
image
Not sure if is something local or general. The dataset is from empiar/11487

oierlauzi:
That is because of the gain function. Try running it without the gain. I dont know why its not being able to read that .gain file

albertmena:
Removing the gain file from importMovie the error chages:
image
Any case we should to check the protocol with the gain file imported, maybe other dataset

oierlauzi:
Did you switch branches in core?

oierlauzi:
> Removing the gain file from importMovie the error chages: image Any case we should to check the protocol with the gain file imported, maybe other dataset

The warnings are expected

albertmena (approved):

albertmena (changes requested):
With Gain file it fails:
image

I use an *.eer file and *.gain from: https://www.ebi.ac.uk/empiar/EMPIAR-11366/

albertmena (approved):
Congrats for the great work!

PR #771, opened 2024-02-06
New protocol: Movie alignment consensus

danielmarchan3:
New protocol for the comparison of two global alignments


albertmena (changes requested):
The test fails
```  File "/home/agarcia/anaconda3/envs/scipion3/lib/python3.8/site-packages/pwem/protocols/protocol_align_movies.py", line 142, in createOutputStep
    raise Exception("All movies failed, didn't create outputMicrographs."
Exception: All movies failed, didn't create outputMicrographs.Please review movie processing steps above.
Protocol failed: All movies failed, didn't create outputMicrographs.Please review movie processing steps above.
```

danielmarchan3:
I checked again and ran it with all my repositories updated, and the two tests passed without any problem. Oier, could you please check where the test is failing? Maybe it has something to do with an old version of the Scipion repositories.

albertmena:
> I checked again and ran it with all my repositories updated, and the two tests passed without any problem. @oierlauzi, could you please check where the test is failing? Maybe it has something to do with an old version of the Scipion repositories.

Now, it works for me

albertmena (approved):

PR #796, opened 2024-06-03
New protocol: PCA classification (streaming and static version) 

danielmarchan3:
New 2D Classification method that can works in streaming and in static mode depending on the protocol you choose. We submitted two protocols that are in BETA phase. 

Eventually, the two protocols can be join in one, however due to their very different purposes we decided to keep them separated for the BETA phase. 

Do not merge before:
https://github.com/I2PC/xmipp/pull/812

https://github.com/I2PC/xmipp/pull/885


albertmena (changes requested):
The test fails:
test_protocols_xmipp_2d.py", line 1550, in checkResults
    self.assertSetSize(prot.outputClasses, size, msg)
AttributeError: 'XmippProtClassifyPca' object has no attribute 'outputClasses'

albertmena (approved):

PR #806, opened 2024-08-12
Locally adjust Particle's CTF according to the micrograph's tilt angle

oierlauzi:
Correct local CTF to account for a tilt angle in the acquisition. Useful for a niche acquisition method where the sample is tilted to avoid preferential orientations.


Vilax (changes requested):
Thanks Oier for this simple but useful protocol. A couple of comments.

It should be possible to choose the tilt axis, that depends on the facility. In tomo the convention is the Y axis, but once the tilt series are aligned. In this case the micrographs lack of alignment. 

The other issue is if the defocus decrease or increase with the z-axis (it implies to add a second flag) that changes the sign of dy.

sorry...

oierlauzi:
The tilt axis can be chosen in the form. The second issue is highlighted in the tilt angle's help, as it can be tweaked with its sign.

Vilax (approved):
Sorry Oier, my bad. I'm not the best reading code ;)

oierlauzi:
I have made the sign parameter explicit

PR #807, opened 2024-09-06
Bug fix in the validate function

cossorzano:


oierlauzi:
The modified line is correct. `validateDLtoolkit` should return a list. In order to append those errors to the existing list of errors, the `+=` operator or `extend` method should be used. If `validateDLtoolkit` is returning something other than a list, this is what should be addressed.

albertmena (approved):

PR #808, opened 2024-09-10
Improve build action

MartinSalinas98:
- Removed redundant comments
- Upgraded external CUDA action to master (no need for future updates)
- Upgraded external checkout action (no need for future updates)
- Used external action for Miniconda installation
- Removed `-noXmipp` flag from Scipion installer (flag seems to have been removed from installer in latest release, and gets interpreted as a conda flag instead: `-noXmipp` --> `-n oXmipp`, creating env `oXmipp`)


Ratolon (approved):
oXmipp jajaja

PR #810, opened 2024-10-02
Fi create gallery update

fede-pe:
Allow Scipion protocol create gallery to access missing projection options from xmipp_phantom_project program.

No changes in xmipp program, just in the protocol so this PR can be merged right away

Test passing


oierlauzi (approved):
Very well documented and implemented

PR #812, opened 2024-10-09
Do not install dependencies in conda by default

albertmena:
https://github.com/I2PC/xmipp/issues/917
Also changed the Action, Holi @MartinSalinas98

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,11 @@ jobs:
       with:
         ref: ${{ github.head_ref }}
 
+    - name: Install dependencies 
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libfftw3-dev libopenmpi-dev openmpi-bin libhdf5-dev python3-numpy python3-dev libtiff5-dev unzip libsqlite3-dev default-jdk git cmake libopencv-dev libopenmpi-dev make cmake
+        
     - name: Install CUDA
       uses: Jimver/cuda-toolkit@master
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -18,6 +18,11 @@ jobs:
       with:
         ref: ${{ github.head_ref }}
 
+    - name: Install dependencies 
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libfftw3-dev libopenmpi-dev openmpi-bin libhdf5-dev python3-numpy python3-dev libtiff5-dev unzip libsqlite3-dev default-jdk git cmake libopencv-dev libopenmpi-dev make cmake
```
MartinSalinas98, line 24: Review this line:
- Didn't we deprecate opencv (cross check with @oierlauzi)
- If this is to be kept, maybe separate Python & openmpi into separate external actions for robustness.

MartinSalinas98:
Maybe this should be discussed with @oierlauzi, since he is the one who implemented the default dependency being handled by conda, to see if it's actually problematic and should be removed as default or it can be fixed.

If to be removed, README.md file should be updated too to reflect this changes in the installation process.

oierlauzi:
I was in person with Alberto when this decision was taken. This is responsible for most of the trouble induced with the new installer

oierlauzi:
I think alberto had updated the README

MartinSalinas98 (approved):

oierlauzi (approved):

Issue #814, opened 2024-10-10
reconstruct highres fails with input SetOfVolumes


jamesmkrieger:
First fnVol1 is not defined. Then when I define it, maxFreq is not defined. This is defined in the else block.

It's also not clear what these input volumes should be. I guess they should be half maps?

PR #821, opened 2024-10-30
JMK fix highres: add unassigned vars fnVol1 fnVol2 to set loop

jamesmkrieger:
I had an error before that these variables were unassigned and I think they should be assigned like this


albertmena:
https://github.com/I2PC/scipion-em-xmipp/issues/814

albertmena (approved):

PR #824, opened 2024-11-15
New protocol: micrographs defocus sampler

danielmarchan3:
This protocol works in streaming. It takes as input a certain number of CTFs and makes a balanced sampling of N images based on the defocus U values. The output will be a sample of CTFs and Micrographs with different defocus values. This kind of sampling can give you a good idea of your defocus distribution and a good subset of variable micrographs with different contrast.

### DIFF
--- a/xmipp3/protocols/protocol_mics_defocus_balancer.py
+++ b/xmipp3/protocols/protocol_mics_defocus_balancer.py
@@ -175,7 +175,7 @@ class XmippProtMicDefocusSampler(ProtCTFMicrographs):
                   %len(self.sampled_images))
 
         stats = compute_statistics(list(ctfDefocus.values()))
-        message = ("The defocus statistics are the following: range %d min %d max %d mean %d std %.1f"
+        message = ("The defocus statistics are the following: range %d   min %d   max %d   mean %d   std %.1f"
                    %(stats["range"], stats["min"], stats["max"],stats["mean"], stats["std"]))
         self.summaryVar.set(message)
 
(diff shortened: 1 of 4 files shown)


xmipp3/protocols/protocol_mics_defocus_balancer.py
```diff
@@ -0,0 +1,296 @@
...
+        sizeOutput = 0
+
+        if hasattr(self, OUTPUT_CTF):
+            sizeOutput = self.outputCTF.getSize()
+            doneIds.extend(list(self.outputCTF.getIdSet()))
+
+        return doneIds, sizeOutput
+
+    def _summary(self):
+        pass
```
albertmena, line 248: Add a small summary, some statistics; min max defocus, mean, bin ranges

xmipp3/protocols/protocol_mics_defocus_balancer.py
```diff
@@ -0,0 +1,296 @@
...
+    def extractBalancedDefocus(self, ctfIds):
+        inputCtfSet = self._loadInputCtfSet(self.ctfFn)
+        ctfDefocus = {}
+
+        for ctfId in ctfIds:
+            ctf = inputCtfSet.getItem("id", ctfId).clone()
+            defocusU = ctf.getDefocusU()
+            ctfDefocus[ctfId] = defocusU
+
+        self.sampled_images = balanced_sampling(image_dict=ctfDefocus, N=self.numImages.get(), bins=10)
```
albertmena, line 173: The bins number should be dynamic based on the number of elements: (Sturges formule)
int(round(1 + math.log2(numElementes))) but forcing bin=3 if lower
That way for bigers sets, more bins and more outputs. 
Consider that option

albertmena (changes requested):
Consider the dynamic bins number and populate the summary

albertmena (approved):

PR #840, opened 2024-12-02
Add GPU use for local angle assignment in HighRes

Ratolon:
Incorporate CZ development on the gpu versio of angular assignment in local highres. Tested with real data.


oierlauzi (approved):
I am not a big fan of passing `numberOfMpi` as threads argument but it is the most reasonable way to achieve parallelization without adding extra complexity for the user.

PR #852, opened 2024-12-11
Add EER support in Dose Analysis

Ratolon:
Xmipp3 Dose Analysis fails when using EER. This is against our latest effort for EER support in Xmipp. Thus, I am implementing EER support for dose analysis protocol.

Bsos,
Mikel.


albertmena:
Related: https://github.com/I2PC/xmipp/issues/939

albertmena (changes requested):
The protocol never ends (2h) with EER files, running 8 files from https://www.ebi.ac.uk/empiar/EMPIAR-11487/

albertmena (commented):
The protocol never ends (2h) with EER files, running 8 files from https://www.ebi.ac.uk/empiar/EMPIAR-11487/

Ratolon:
This protocol clearly needs more work so let's keep it a draft

PR #866, opened 2024-12-18
managin scipionApp

albertmena:
Check if scipion-app <= 3.7.0 to avoid the installation because of the [ scipion-app bug ](https://github.com/scipion-em/scipion-app/pull/201)

### DIFF
--- a/xmipp3/__init__.py
+++ b/xmipp3/__init__.py
@@ -170,8 +170,8 @@ class Plugin(pwem.Plugin):
             )
         
         #Release binaries installation
-	vScipionApp = Version(scipionAppVersion)
-	vMinRequired = Version('3.7.1')
+        vScipionApp = Version(scipionAppVersion)
+        vMinRequired = Version('3.7.1')
         if vScipionApp < vMinRequired:
 	        print("\n---------------------------------------------------\nAttention! The current version of 'scipion-app' is outdated.\nTo update it to the latest version, please run the following command in your terminal:\n\n  scipion3 update\n\nThis command will update 'scipion-app' to the latest available version.\nNote: The minimum required version of 'scipion-app' is 3.7.1.")
         else:
--- a/xmipp3/__init__.py
+++ b/xmipp3/__init__.py
@@ -173,7 +173,7 @@ class Plugin(pwem.Plugin):
 	vScipionApp = Version(scipionAppVersion)
 	vMinRequired = Version('3.7.1')
         if vScipionApp < vMinRequired:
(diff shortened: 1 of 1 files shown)


xmipp3/__init__.py
```diff
@@ -167,26 +168,30 @@ def defineBinaries(cls, env):
...
-            neededProgs=['git', 'gcc', 'g++', 'cmake', 'make'],
-            updateCuda=True,
-            default=not develMode
-        )
-
-        ## EXTRA PACKAGES ##
-        installDeepLearningToolkit(cls, env)
-
+        #Release binaries installation
+        scipionApp = int(''.join(map(str, [int(x) for x in scipionAppVersion.split('.')])))
```
MartinSalinas98, line 172: Overcomplicated. You do:
1. split version string by dots (you get a list of strings)
2. do a for comprehension to convert the list of strings into a list of ints
3. do a map on that list of ints converting them back to a list of strings
4. join them to make them a single string
5. convert the result number string into an int

Why not just do:
```py
int(scipionAppVersion.replace(".", ""))
```
1. Remove dots from the version string
2. convert into int

Still this is not a foolproof method:
Let's say we require scipion-app version `3.7.0`. The methods above would convert it into `370`.
If tomorrow scipion releases scipion `3.8.0`, it would be translated into `380`, which works as intended.
But version `3.6.10` is prior to `3.7.0` in semver's system, however, it gets converted into `3610`, considered greater, and therefore accepted.

MartinSalinas98: TLDR: use a comparison function:
```py
def meetsRequiredVersion(version, requiredVersion):
    requiredVersionList = requiredVersion.split(".")
    versionList = version.split(".")
    if len(versionList) != len(requiredVersionList):
        # Show error, the version formats do not match
        return False # Or even raise exception
    for versionSegment, minimumVersionSegment in zip(versionList, requiredVersionList):
        if versionSegment < minimumVersionSegment:
            print("Version minimum than required.....")
            return False
    return True
```

MartinSalinas98 (approved):

MartinSalinas98 (changes requested):
Approved by mistake

xmipp3/__init__.py
pconesa: This migth be useful: https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

MartinSalinas98: > This migth be useful: https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

I agree, that's even better, didn't know the standard library came with a version comparator built in

albertmena: Great, thanks both, I included the packaging.version

PR #877, opened 2025-01-08
Protocol dose analysis: make the estimation tolerable to corrupted movies files

danielmarchan3:
If it encounters a movie that is corrupted or it is not a movie, it will avoid this file and continue.


albertmena (changes requested):
Please add details about the protocol in the help.
L454: "Dose (electrons impacts per angstrom**2 )" -> "Dose ($\e^-$ impacts per $\AA²$)"
colors plot: upper and lower limit with the same color (no red no green, those use to means good/bad), median change too

albertmena (approved):

PR #883, opened 2025-01-21
init generator for pre and post ctf noise

### LINKED ISSUE #882: Simulating CTF with pre-CTF noise doesn't work
First I had a problem that I fixed in #881. Now there's another one:

00001:   Traceback (most recent call last):
00002:     File "/home/spa/scipion3/scipion-pyworkflow/pyworkflow/protocol/protocol.py", line 207, in run
00003:       self._run()
00004:     File "/home/spa/scipion3/scipion-pyworkflow/pyworkflow/protocol/protocol.py", line 257, in _run
00005:       resultFiles = self._runFunc()
00006:     File "/home/spa/scipion3/scipion-pyworkflow/pyworkflow/protocol/protocol.py", line 253, in _runFunc
00007:       return self._func(*self._args)
00008:     File "/home/spa/scipion3/xmipp-bundle/src/scipion-em-xmipp/xmipp3/protocols/protocol_simulate_ctf.py", line 124, in simulateStep
00009:       I.setData(Idata+self.noiseBefore.get()*np.random.Generator.normal(size=Idata.shape))
00010:   TypeError: descriptor 'normal' of 'numpy.random._generator.Generator' object needs an argument
00011:   Protocol failed: descriptor 'normal' of 'numpy.random._generator.Generator' object needs an argument

jamesmkrieger:
Fixes #882

### DIFF
--- a/xmipp3/protocols/protocol_simulate_ctf.py
+++ b/xmipp3/protocols/protocol_simulate_ctf.py
@@ -121,7 +121,8 @@ class XmippProtSimulateCTF(Prot2D):
             if self.noiseBefore>0:
                 I=xmippLib.Image(fnIn)
                 Idata = I.getData()
-                I.setData(Idata+self.noiseBefore.get()*np.random.Generator.normal(size=Idata.shape))
+                generator = np.random.default_rng()
+                I.setData(Idata+self.noiseBefore.get()*generator.normal(size=Idata.shape))
                 I.write(fnOut)
                 fnIn=fnOut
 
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
A potential sanity test from my log likelihood analysis:

I simulated projections from 2 volumes converted from pdbs of the spike in different conformations and am looking at how the sums of square residuals, which are the simplest part of the log likelihood equation, are affected by various operations.

- when I normalise synthetic projections after simulating the CTF without noise, I get sums of square residuals from reprojections of two volumes that are rather even. They have small differences that can be captured by normalisation by the mean over volumes
- when I add noise with a separate add noise protocol using Gaussian noise with standard deviation 1.0 before simulating CTF and when I use the pre-CTF noise option here with sigma 1.0, I get the same behaviour that normalising the images does not flatten the sums of square residuals.

Simulate CTF + preprocess norm, no noise
image

Add noise + simulate CTF (separate protocols) + preprocess norm
image

Simulate CTF with pre-CTF (same protocol) + preprocess norm
image

albertmena (approved):

PR #887, opened 2025-01-30
Fix FSCQ test

DavidHerreros:
FSCQ test was not passing due to a wrong conversion done by Chimera (import pdb from file -> PDB -> CIF). 

Instead of relying on a local file, we can import directly from the database to avoid the conversion.


albertmena:
But the error remains:
```
terminate called after throwing an instance of 'cif::validation_error'
  what():  When validating _entity.pdbx_description: Value 'hemoglobin subunit α' does not match type expression for type line
```

albertmena (changes requested):
The error remains

PR #892, opened 2025-02-24
deepLearningToolkit working with Apptainer

sanlopez:
The installation of deepLearningToolkit creates several conda environments such as "xmipp_DLTK_v0.3", which include the tensorflow python module. The way it checks if tensorflow should be installed in CPU or GPU mode is by executing "nvidia-smi". The issue is that when installing it through Appatainer, at the time of building the container, you don't have access to the GPU host resources, so you can't execute "nvidia-smi". Instead, GPU access becomes available later, when you actually run the previously built container. Therefore, this is an alternative way to check if the host machine has a GPU.

### DIFF
--- a/xmipp3/__init__.py
+++ b/xmipp3/__init__.py
@@ -216,6 +216,22 @@ class Plugin(pwem.Plugin):
                     os.path.isfile(os.path.join(bundleDir, 'xmipp')))
         
         return bundleDir if isBundle else None
+
+def getNvidiaDriverVersion(plugin):
+    """ Several ways to retrieve NVIDIA driver version.
+    """
+    commands = [["nvidia-smi", "--query-gpu=driver_version", "--format=csv,noheader"],
+                ["cat", "/sys/module/nvidia/version"]]
+    for cmd in commands:
+        try:
+            nvidiaDriverVer = subprocess.Popen(cmd,
+                                               env=plugin.getEnviron(),
+                                               stdout=subprocess.PIPE
+                                               ).stdout.read().decode('utf-8').split(".")[0]
+            return nvidiaDriverVer
+        except (ValueError, TypeError, FileNotFoundError):
+            continue
+    return None
     
 def installDeepLearningToolkit(plugin, env):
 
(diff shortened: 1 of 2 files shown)


xmipp3/__init__.py
```diff
@@ -216,45 +216,55 @@ def __getBundleDirectory(cls):
...
+    for cmd in commands:
+        try:
+            nvidiaDriverVer = subprocess.Popen(cmd,
+                                               env=plugin.getEnviron(),
+                                               stdout=subprocess.PIPE
+                                               ).stdout.read().decode('utf-8').split(".")[0]
+            return nvidiaDriverVer
+        except (ValueError, TypeError, FileNotFoundError):
+            continue
+    return None
```
MartinSalinas98, line 234: This line is redundant

MartinSalinas98 (changes requested):

xmipp3/__init__.py
```diff
@@ -216,45 +216,55 @@ def __getBundleDirectory(cls):
                     os.path.isfile(os.path.join(bundleDir, 'xmipp')))
         
         return bundleDir if isBundle else None
+
+def getNvidiaDriverVersion(plugin):
+    """ Several ways to retrieve NVIDIA driver version.
+    """
+    commands = [["nvidia-smi", "--query-gpu=driver_version", "--format=csv,noheader"],
+                ["cat", "/sys/module/nvidia/version"]]
```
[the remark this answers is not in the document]
sanlopez: The version of the nvidia driver. And yes, with `nvidia-smi --query-gpu=driver_version --format=csv,noheader
` you get the same as with `cat /sys/module/nvidia/version`

MartinSalinas98:
For me, after removing line 234 (`return None` is equivalent to not having a return statement at all) looks good to be merged.

albertmena (approved):

xmipp3/__init__.py
MartinSalinas98:
```py
def my_func(my_string: str):
    if my_string == "YES":
        return mystring
    return None

def my_func(my_string: str):
    if my_string == "YES":
        return mystring
    return

def my_func(my_string: str):
    if my_string == "YES":
        return my_string
```
Those three functions do the same thing.
So, `return None` and `return` do the same thing anywhere in the code. But, additionally, if the return statement is the last line of the function, then, also not having a return statement there acts the same as not returning anything (therefore, returning `None`)

So, in your case, you could remove that `return None` line and the function would still work exactly the same way.

sanlopez: Done! Thank you, Martín :)

albertmena: Enjoyed the talk!

PR #902, opened 2025-03-10
better label for volume 1 vs 2

jamesmkrieger:
It's not useful for users to just see Volume 1 and Volume 2. I had no idea a priori which one was the one that would be modified. I have confirmed against the xmipp program, and this matches what Carlos Oscar said:
```
	addParamsLine("--i1 <volume>			: Reference volume");
	addParamsLine("--i2 <volume>			: Volume to modify");
	addParamsLine("[-o <structure=\"\">]\t: Volume 2 modified or "
			"volume difference");
```


albertmena (approved):
Thanks

PR #903, opened 2025-03-10
JMK_COSS log likelihood for reweighting

jamesmkrieger:
This adds a new protocol and viewer for calculating image-volume log likelihoods using either angular_continuous_assign2 or my new program in xmipp branch [jmk_new_residuals](https://github.com/I2PC/xmipp/tree/jmk_new_residuals) (see https://github.com/I2PC/xmipp/pull/991).

The protocol compares a set of N images against M volumes and outputs a duplicated set of images (NxM) with columns for residuals and log likelihoods. It also outputs a matrix as in npy file format, which is the input for the reweighting estimate protocol and the viewer.

This now works well for real images e.g. 
Screenshot 2025-03-05 124338


xmipp3/protocols/protocol_compute_likelihood.py
```diff
@@ -0,0 +1,288 @@
...
+from pyworkflow.protocol.constants import LEVEL_ADVANCED, STEPS_PARALLEL
+
+
+from xmipp3.convert import setXmippAttributes
+import xmippLib
+
+
+
+class XmippProtComputeLikelihood(ProtAnalysis3D):
+    """This protocol computes the likelihood of a set of particles with assigned angles when compared to a
```
albertmena, line 52: remove the line break

albertmena (changes requested):
Great job, but Xmipp needs a test for all protocols, especially if it is a new one.

xmipp3/protocols/protocol_compute_likelihood.py
jamesmkrieger: Ok

jamesmkrieger:
> Great job, but Xmipp needs a test for all protocols, especially if it is a new one. It also helps me to check the protocol and future changes

Ok, I will make one. Thanks

jamesmkrieger:
Do we have a test data set with 2 volumes and the corresponding particles that I can use?

jamesmkrieger:
or where can I add one?

jamesmkrieger:
> > Do we have a test data set with 2 volumes and the corresponding particles that I can use?
> 
> I found one in the relion tutorial dataset

Actually, the low resolution made that not work, so I will use relion particles with one of those volumes with a resized spike emdb volume

albertmena (approved):
Great!

PR #905, opened 2025-03-12
Use of dynamic dependencies

martin-s-a:
## Note: This PR needs to be merged after [xmipp-995](https://github.com/I2PC/xmipp/pull/995).
- `xmipp3-installer` will now be automatically installed with the plugin to ease Xmipp's new installation process.
- Instead of pointing to a specific Xmipp's version (`v3.24.0`, for example), a new dynamic tag `v3` is used, to avoid needing to release a new version of the plugin just because a new release of xmipp was generated. Now releases in the plugin will only be tiggered by changes in the code of the plugin itself, nothing else.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,5 @@ joblib
 peppercorn
 plotter
 PyWavelets
-xmipp3-installer==1.*
+xmipp3-installer==2.*
 numpy==1.*


xmipp3/version.py
```diff
@@ -28,8 +28,8 @@
 type_of_version = 'devel' #'release' 'devel'
 _logo = "xmipp_logo" + ("" if type_of_version == 'release' else '_devel') + '.png'
 
-_binVersion = 'v3.24.12.0' # Increase it if hotfix in binaries (xmipp, xmippCore and/or XmippViz)
-_pluginVersion = 'v3.24.12.2' # Increase it if hotfix in binaries (xmipp, xmippCore and/or XmippViz) or in scipion-em-xmipp
+_binVersion = 'v3' # Increase it if major release is generated in xmipp
+_pluginVersion = 'v3.25.0' # Increase according to SemVer rules
```
oierlauzi, line 32: I would add the comments regarding the instructions of how to handle `_pluginVersion` bump ups.

oierlauzi (changes requested):
I see that `_binTagVersion` and `_pluginTagVersion` are still around with the release name (Oceanus, Poseidon...). With the new semantic versioning, do they still make sense? Even if we give names to major releases, it is possible that plugin-bin names to get out of sync.

MartinSalinas98:
> I see that `_binTagVersion` and `_pluginTagVersion` are still around with the release name (Oceanus, Poseidon...). With the new semantic versioning, do they still make sense? Even if we give names to major releases, it is possible that plugin-bin names to get out of sync.

Starting from this new release, the naming also will change, in which way still has to be decided, but it's a minor issue, since it's only something "visual". There are several options:
- Remove the release names completely, leaving only the version numbers. The easiest to do, might release names have some marketing & sentimental value (basically they are kind of cool, but that's all).
- Leave the release names and make a new one for each new release (impractical, as with the new system releases could be generated as quick as in a weekly basis).
- Only add release names to major releases, leaving that name unchanged for all the minor releases of such major. This can cause the name desync between repositories that you mentioned (although I'm not sure that's actually a problem in itself, as the naming convention for the different repositories could just diverge and follow different paths with even different names for the same letter at different points in time), and can also make some names last longer than others, as breaking changes in the codebase can and will occur without an evenly spread development time between them.

In my opinion, we could remove them altogether, as it is the simpler and more trouble-free option, but that's something that requires certain degree of consensus to happen.

xmipp3/version.py
MartinSalinas98: I added the text `SemVer rules` so those rules can be searched for externally, but, if you think it's necessary, I can add a project-specific set of rules as to when bump up which version number.

albertmena:
The init.py has to change to launch the xmipp3_installer

MartinSalinas98:
> The init.py has to change to launch the xmipp3_installer

No, it doesn't, as [xmipp already calls the installer](https://github.com/I2PC/xmipp/blob/ms_python_installer/xmipp#L58).

oierlauzi:
> > I see that `_binTagVersion` and `_pluginTagVersion` are still around with the release name (Oceanus, Poseidon...). With the new semantic versioning, do they still make sense? Even if we give names to major releases, it is possible that plugin-bin names to get out of sync.
> 
> Starting from this new release, the naming also will change, in which way still has to be decided, but it's a minor issue, since it's only something "visual". There are several options:
> 
>     * Remove the release names completely, leaving only the version numbers. The easiest to do, might release names have some marketing & sentimental value (basically they are kind of cool, but that's all).
> 
>     * Leave the release names and make a new one for each new release (impractical, as with the new system releases could be generated as quick as in a weekly basis).
> 
>     * Only add release names to major releases, leaving that name unchanged for all the minor releases of such major. This can cause the name desync between repositories that you mentioned (although I'm not sure that's actually a problem in itself, as the naming convention for the different repositories could just diverge and follow different paths with even different names for the same letter at different points in time), and can also make some names last longer than others, as breaking changes in the codebase can and will occur without an evenly spread development time between them.
> 
> 
> In my opinion, we could remove them altogether, as it is the simpler and more trouble-free option, but that's something that requires certain degree of consensus to happen.

I agree.

Ratolon:
> > I see that `_binTagVersion` and `_pluginTagVersion` are still around with the release name (Oceanus, Poseidon...). With the new semantic versioning, do they still make sense? Even if we give names to major releases, it is possible that plugin-bin names to get out of sync.
> 
> Starting from this new release, the naming also will change, in which way still has to be decided, but it's a minor issue, since it's only something "visual". There are several options:
> 
> * Remove the release names completely, leaving only the version numbers. The easiest to do, might release names have some marketing & sentimental value (basically they are kind of cool, but that's all).
> * Leave the release names and make a new one for each new release (impractical, as with the new system releases could be generated as quick as in a weekly basis).
> * Only add release names to major releases, leaving that name unchanged for all the minor releases of such major. This can cause the name desync between repositories that you mentioned (although I'm not sure that's actually a problem in itself, as the naming convention for the different repositories could just diverge and follow different paths with even different names for the same letter at different points in time), and can also make some names last longer than others, as breaking changes in the codebase can and will occur without an evenly spread development time between them.
> 
> In my opinion, we could remove them altogether, as it is the simpler and more trouble-free option, but that's something that requires certain degree of consensus to happen.

I think this option is good, albeit with a previous discussion to define some minor details. It could get us more time to focus on what really matters.

Wbw..M

albertmena:
I am of the opinion to remove the release names.

requirements.txt
```diff
@@ -9,3 +9,4 @@ peppercorn
 plotter
 PyWavelets
 xmipp3-installer==1.*
+numpy==2.*
```
albertmena: Scipion has a numpy 1.24 and scipion-em-xmipp will be always with Scipion. 
Why numpy 2.0?

albertmena (changes requested):

requirements.txt
martin-s-a: It was arbitrarily a recent version. I think it also works with 1.*

albertmena (approved):

PR #912, opened 2025-03-31
protocol_ctf_consensus: Changes in functionality

EduGitt:
Some changes were made due to some malfunction regarding the termination of the protocol when having two sets of micrographs/CTFs with different IDs, and not selecting correctly a common set of IDs. That caused the program to loop itself and never end. That problem is now solved.


EduGitt:
Also, its test has been run and it works perfectly with the changes applied.

PR #915, opened 2025-04-01
Generate release based on Scipion standard

albertmena:


MartinSalinas98 (changes requested):

MartinSalinas98:
In summary, i wouldn't use sicpion-em's reusable workflow for xmipp's release, at least for now, since it's lacking some features we would need:

- No release creation (with our custom way of parsing the changelog to automatically add it in the release)
- No output artifacts generated. Their workflow could implement the upload of some artifacts with the built project & some metadata to be consumed by extra steps outside. This could include tag version/name created, which then could be used by a next step that could add the missing details, but it's not possible without the artifacts and without basically redoing most of the steps, defeating the point of the reusable workflow.

For now, i'd leave it like that, some things can be improved about the current setup, but i don't think the reusable workflow meets the needs of this repository's release system just yet.

albertmena:
ocai

albertmena:
Is neccesary to read from the CHANGELOG to get the version? It is available on the xmipp3/version.py

MartinSalinas98:
> Is neccesary to read from the CHANGELOG to get the version? It is available on the xmipp3/version.py

Not really, you are right about that, we only really need the changelog for the chengelog text itself, and it's likely safer to get the version from the python code. We should change that. Doing:
```
export VERSION=$(python -c "import importlib.metadata; print(importlib.metadata.version(\"xmipp3\"))")
```
Should retrieve it properly i believe.

albertmena:
Ok. The next release will be the last  without the xmipp3-installer, and the last with the gods naming. So the changes in the version parsing are not applied yet

PR #910, opened 2025-03-27
protocol_movie_alignment_consensus: Changes in functionality

EduGitt:
Changes in functionality of Movie Alignment Consensus within comparisons between ons between shifts trajectories, and a additional change related to the minimum consensus or threshold for shifts correlation.

Some code lines have been added so that the consensus between shifts trajectories could be better and could be easily compared. Also, some factor related to the consensus threshold has been added.

PR #919, opened 2025-04-04
protocol_particle_pick_consensus: Changes in functionality and errors

EduGitt:
Some changes were made in this protocol due to some malfunctions regarding the selection and indexing of some variables by IDs. The lengths of each set of coordinates provided by a certain number of micrograph were different and their IDs were too, adding to problems such as "'NoneType' object has no attribute 'X'", being X an specific attribute, when indexing by wrong IDs.


EduGitt:
Sorry for the additional commits that doesn't belong to this pull request. Only the last one is valid.

albertmena:
> Sorry for the additional commits that doesn't belong to this pull request. Only the last one is valid.

But this commits should not be here, because those are being reviewed here: https://github.com/I2PC/scipion-em-xmipp/pull/910/files/604942b4501ceef1aedbbdeb448bb0a1a60dc796..f417df99c7b5f806c804baa06ed66e0dfca567b9
Lets close this  PR: #919 and create new one with just the commits related to the particle_pick_consensus

PR #921, opened 2025-04-10
protocol_particle_pick_consensus: Changes in functionality

EduGitt:
When selecting and picking particles from two different sets of coordinates, there were some errors regarding the correct use of IDs from each set when their length and IDs were different from each other. That led to discrepancies such as "'NoneType' object has no attribute 'X' ", being X an specific attribute, due to some variables appearing when indexing by ID as 'None'. Now, that is corrected, and the protocol now can be used also for assymetrical sets.

### DIFF
--- a/xmipp3/protocols/protocol_particle_pick_consensus.py
+++ b/xmipp3/protocols/protocol_particle_pick_consensus.py
@@ -169,6 +169,11 @@ class XmippProtConsensusPicking(ProtParticlePicking):
                 readyMics.intersection_update(currentPickMics)
             allMics = allMics.union(currentPickMics)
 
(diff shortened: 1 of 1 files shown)


albertmena (changes requested):
Replace not == by !=
You can visit this web that review the code and warns about potential bugs:

https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=921&id=I2PC_scipion-em-xmipp&open=AZYgeZ1DT5X3jBavkeFa

albertmena (approved):

PR #923, opened 2025-04-23
protocol_movie_split_frames and test_protocol_split_frames: Changes in functionality

EduGitt:
Split Frames protocol has been modified, to the extent that its configuration is different from before: now it allows to prove if the sum_frames method is necessary when creating micrographs from the input movies, it now has a summary, improvement on its layout... And therefore, its corresponding test is now adapted to the new changes.

### DIFF
--- a/xmipp3/protocols/protocol_movie_split_frames.py
+++ b/xmipp3/protocols/protocol_movie_split_frames.py
@@ -1,6 +1,7 @@
 # **************************************************************************
 # *
 # * Authors:     Jose Luis Vilas Prieto (jlvilas@cnb.csic.es)
+# *              Eduardo García Delgado (eduardo.garcia@cnb.csic.es)
 # *
 # * Unidad de  Bioinformatica of Centro Nacional de Biotecnologia , CSIC
 # *
@@ -32,7 +33,7 @@ from pyworkflow.protocol.params import (PointerParam, BooleanParam)
 
 from pwem.protocols import ProtPreprocessMicrographs
 
-from pwem.objects import Movie, Micrograph, SetOfMovies
+from pwem.objects import Movie, Micrograph
 
 
 class XmippProtSplitFrames(ProtPreprocessMicrographs):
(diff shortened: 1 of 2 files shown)


xmipp3/protocols/protocol_movie_split_frames.py
```diff
@@ -53,16 +53,9 @@ def _defineParams(self, form):
...
-        group.addParam('Micrographs', BooleanParam,
-                       label="Micrographs Selection", important=False, default=False,
-                       help='Set *Yes* to select a set of micrographs just in case alignment has been processed '
-                            'in the input movies, so that they can also be split into two sets (odd  and even) of micrographs. '
-                            'Set *No* if alignment has not been processed in the input movies.')
-
-        group.addParam('inputMicrographs', PointerParam, pointerClass='SetOfMicrographs',
-                      label="Input micrographs", condition="Micrographs",
-                      help='Select a set of micrographs from the previous protocol.')
+        form.addParam('sumFrames', BooleanParam,
```
albertmena, line 56: For a more comprensive menu please consider the EnumParam to provide a list of options (movies, micrographs). Something like:
```
form.addParam('sumFrames', EnumParam,  label="Sum Frames", important=False, default=0,
choices=["moviesl", "micrographs"],
help='Set movies or micrographs to get one of those"')
```

albertmena (changes requested):

xmipp3/protocols/protocol_movie_split_frames.py
EduGitt: But in this case, as you can see, I changed the input micrographs option for a boolean one to provide an easier way to provide an output for micrographs, and because I realized that didn't imply alignment necessarily. The input one was too much of a stretch to assess that issue. So now, the only necessary input is the one from the incoming movies.

albertmena: Sum Frames = YES meaning get a set of micrograms
Sum Frames = NO  meaning get a set of movies

What I want to propose is make it more readable:
`Sum frames: set of movies / set of micropgraphs`
This can be done with an EnumParam as I showed

albertmena (changes requested):
We can see it in person too

albertmena:
The test has to change, sumFrames does not exist

albertmena (changes requested):
change test parameters

albertmena (approved):
Great job!

PR #935, opened 2025-06-09
Protocol Structure map: Adding a dendrogram 

cossorzano:
Adding a dendrogram to the structMap viewer and cleaning many intermediate files.

### DIFF
--- a/xmipp3/viewers/viewer_structure_map.py
+++ b/xmipp3/viewers/viewer_structure_map.py
@@ -122,14 +122,13 @@ class XmippProtStructureMapViewer(ProtocolViewer):
         condensed_dist = squareform(0.5*(distMatrix+distMatrix.transpose()))
         Z = linkage(condensed_dist, method='complete')
 
-        fig, ax = plt.subplots()
-        dendrogram(Z, ax=ax)
-        ax.set_title("Hierarchical Clustering (Complete Linkage)")
-        ax.set_xlabel("Sample Index")
-        ax.set_ylabel("Distance")
-        fig.tight_layout()
-
-        return [fig]
+        plt.figure(figsize=(8, 4))
+        dendrogram(Z)
+        plt.title("Hierarchical Clustering (Complete Linkage)")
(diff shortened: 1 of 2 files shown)


xmipp3/viewers/viewer_structure_map.py
```diff
@@ -112,6 +117,19 @@ def _visualize(self, e=None):
...
+        condensed_dist = squareform(0.5*(distMatrix+distMatrix.transpose()))
+        Z = linkage(condensed_dist, method='complete')
+
+        plt.figure(figsize=(8, 4))
+        dendrogram(Z)
+        plt.title("Hierarchical Clustering (Complete Linkage)")
+        plt.xlabel("Sample Index")
+        plt.ylabel("Distance")
+        plt.tight_layout()
+        plt.show()
```
oierlauzi, line 131: Avoid using `plt.show()` as it blocks the whole Scipion execution thread

```
fig, ax = plt.subplots()
dendrogram(Z, ax=ax)
ax.set_title("Hierarchical Clustering (Complete Linkage)")
ax.set_xlabel("Sample Index")
ax.set_ylabel("Distance")
fig.tight_layout()

return [fig]
```

DavidHerreros (approved):

xmipp3/viewers/viewer_structure_map.py
albertmena: With this change the action fails, can we revert it? or have you any idea about the error? @oierlauzi https://github.com/I2PC/scipion-em-xmipp/actions/runs/15698553546/job/44325017124

albertmena:
@oierlauzi @DavidHerreros The action fails and just this PR fails. 
I tried with the proposal of Oier in the viewer and without it, both fails.

oierlauzi:
> @oierlauzi @DavidHerreros The action fails and just this PR fails. I tried with the proposal of Oier in the viewer and without it, both fails.

I don't think the viewer was the problem because it involves python code that has nothing to do with the installer. Some dependency or the GH worker may have changed.

PR #937, opened 2025-06-10
deep_hand: erasing cuda things from env['LD_LIBRARY_PATH']

albertmena:


oierlauzi:
Have you tested it in a setup with both pathologies? (The one that appears with and without LD_LIBRARY_PATH='')

oierlauzi:
I would not merge until we tested the other case. I will try to replicate

albertmena:
> I would not merge until we tested the other case. I will try to replicate

Ok it was tested on @milgrom machine (thanks @sanlopez )

PR #950, opened 2025-06-25
fixing validation

albertmena:


xmipp3/protocols/protocol_deep_hand.py
```diff
@@ -177,6 +177,8 @@ def _methods(self):
     def _validate(self):
         errors = []
 
+        errors.append(self.validateDLtoolkit())
```
oierlauzi, line 180: errors += self.validateDLtoolkit()

PR #955, opened 2025-06-26
Fixing CTF estimation error when downsampling is set to 1

oierlauzi:
Closes https://github.com/I2PC/xmipp/issues/1018


albertmena (approved):
Works for me too

PR #957, opened 2025-06-26
Simplified cuda management 

albertmena:
Handling CUDA_VISIBLE_DEVICES

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,10 +18,10 @@ jobs:
       with:
         ref: ${{ github.head_ref }}
 
-    - name: Install dependencies 
+    - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y libopenmpi-dev openmpi-bin python3-numpy git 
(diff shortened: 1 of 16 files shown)


xmipp3/protocols/protocol_screen_deepConsensus.py
```diff
@@ -487,14 +487,10 @@ def _insertAllSteps(self):
...
     def setGPU(self):
-        if self.useQueueForSteps() or self.useQueue():
-            myStr = os.environ["CUDA_VISIBLE_DEVICES"]
-        else:
-            myStr = self.gpuList.get()
-            os.environ["CUDA_VISIBLE_DEVICES"] = self.gpuList.get()
-        self.numGPU = myStr.split(',')[0]
-
-
+        self.protGpus = " ".join(map(str, self._stepsExecutor.getGpuList()))
```
danielmarchan3, line 490: This is not correct, since CUDA_VISIBLE_DEVICES expect a string separated with commas. Have a look at highres implementation.

danielmarchan3 (changes requested):
Have a look at high res implementation. Setting CUDA_VISIBLE_DEVICES requires a string separated with commas: "0, 1, 2, etc."

danielmarchan3 (changes requested):
Remove all self.protGpus lines, since we are not using that variable anymore.

PR #966, opened 2025-08-04
Disable xmippDep by default as it is breaking installations

oierlauzi:


oierlauzi:
For the future, the path of least damage seems to only install `libstdcxx-ng` in conda and require the user to install everything else.

PR #968, opened 2025-08-25
Change the criteria to evaluate the shift

danielmarchan3:
- Changed frame shift evaluation to use global vector magnitudes instead of per-axis maxima.
- For global shift, replaced net displacement calculation with accumulated path length over all frames to better capture non-linear motion.
- Updated default thresholds to 10 Å (max frame shift) and 45 Å (max movie shift) based on empirical testing for similar rejection rates as the previous method.

### DIFF
--- a/xmipp3/protocols/protocol_movie_max_shift.py
+++ b/xmipp3/protocols/protocol_movie_max_shift.py
@@ -106,13 +106,13 @@ class XmippProtMovieMaxShift(ProtProcessMovies):
                            ' - *by frame or movie*: Rejects movies if one of '
                                     'the conditions above are met.')
 
-        form.addParam('maxFrameShift', params.FloatParam, default=5,
+        form.addParam('maxFrameShift', params.FloatParam, default=10,
                        label='Max. frame shift (A)',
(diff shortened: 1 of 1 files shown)


xmipp3/protocols/protocol_movie_max_shift.py
```diff
@@ -234,24 +234,30 @@ def _evaluateMovieAlign(self, movIds):
...
                 shiftArrayY = np.asarray(shiftListY)
 
                 evalBoth = self.rejType==self.REJ_AND or self.rejType==self.REJ_OR
-                if self.rejType == self.REJ_MOVIE or evalBoth:
-                    rangeX = np.max(shiftArrayX) - np.min(shiftArrayX)
-                    rangeY = np.max(shiftArrayY) - np.min(shiftArrayY)
-                    rangeM = max(rangeX, rangeY) * sampling
-                    rejectedByMovie = rangeM > self.maxMovieShift.get()
 
+                # --- Evaluacion por desplazamiento acumulado ---
```
albertmena, line 245: comments in English

danielmarchan3:
Test and PRs comments updated, ready to merge.

PR #1021, opened 2026-05-28
Fix TestXmippScreenParticles

DavidHerreros:
Correction of a typo when setting a form parameter in the test leading to a bug in the test execution.


albertmena (approved):
Quick and easy—great!

